PR #9, opened 2017-08-03
Add polyline struct

### LINKED ISSUE #2: Polyline struct
Assuming we aren’t overly concerned about Objective-C compatibility, we should encapsulate the concept of a “polygon” in a struct named `Polyline` or `Line` (to use Turf.js terminology), instead of assuming that any array of `CLLocationCoordinate2D`s is intended to be connected sequentially. We could then place most of the freestanding functions under `Polyline` as instance methods.

Is there a more Swiftian representation than a struct that would allow client code to take advantage of all the Array methods more directly?

/cc @frederoni @bsudekum

frederoni:
Fixes #2 - encapsulate polyline methods in a Polyline struct.

@bsudekum @1ec5 👀

### DIFF
--- a/Turf/Turf.swift
+++ b/Turf/Turf.swift
@@ -118,14 +118,26 @@ public struct Turf {
     }
 }
 
+/**
+ A `Polyline` struct represents a shape consisting of two or more coordinates,
+ specified as [CLLocationCoordinate2D]
+ */
 public struct Polyline {
     
+    /**
+     `IndexedCoordinate` is a coordinate with additional information such as
+     the index from its position in the polyline and distance from the start
+     of the polyline.
+     */
     public struct IndexedCoordinate {
         let coordinate: Array<CLLocationCoordinate2D>.Element
         let index: Array<CLLocationCoordinate2D>.Index
         let distance: CLLocationDistance
     }
     
+    /**
+     The coordinates that the `Polyline` was initialized with.
+     */
     public var coordinates: [CLLocationCoordinate2D]
     
     /**
--- a/Turf/Turf.swift
+++ b/Turf/Turf.swift
@@ -140,7 +140,7 @@ public struct Polyline {
     /**
      Returns a coordinate along a polyline at a certain distance from the start of the polyline.
      */
-    public func coordinateFromStart(at distance: CLLocationDistance) -> CLLocationCoordinate2D? {
+    public func coordinateFromStart(distance: CLLocationDistance) -> CLLocationCoordinate2D? {
         // Ported from https://github.com/Turfjs/turf/blob/142e137ce0c758e2825a260ab32b24db0aa19439/packages/turf-along/index.js
         var traveled: CLLocationDistance = 0
         
--- a/Turf/Turf.swift
+++ b/Turf/Turf.swift
@@ -120,7 +120,7 @@ public struct Turf {
 
 /**
  A `Polyline` struct represents a shape consisting of two or more coordinates,
- specified as [CLLocationCoordinate2D]
+ specified as `[CLLocationCoordinate2D]`
  */
 public struct Polyline {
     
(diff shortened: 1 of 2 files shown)


Turf/Turf.swift
```diff
@@ -107,126 +107,117 @@ public struct Turf {
...
+        return coords
+    }
+    
+    
     /**
      Returns a coordinate along a polyline with x units away from a coordinate.
      */
-    public static func polyline(along polyline: [CLLocationCoordinate2D], within distance: CLLocationDistance, of coordinate: CLLocationCoordinate2D) -> [CLLocationCoordinate2D] {
-        let startVertex = closestCoordinate(on: polyline, to: coordinate)
+    public func polyline(within distance: CLLocationDistance, of coordinate: CLLocationCoordinate2D) -> [CLLocationCoordinate2D] {
```
1ec5: `trimmed(from:distance:)`

1ec5: This method should also return a Polyline.

Turf/Turf.swift
```diff
@@ -268,19 +259,44 @@ public struct Turf {
...
-    /**
-     Returns the smallest angle between two angles.
+     Returns the geographic coordinate along the polyline that is closest to the given coordinate as the crow flies.
+     
+     The returned coordinate may not correspond to one of the polyline’s vertices, but it always lies along the polyline.
      */
-    public static func differenceBetweenAngles(_ alpha: CLLocationDirection, _ beta: CLLocationDirection) -> CLLocationDirection {
-        let phi = abs(beta - alpha).truncatingRemainder(dividingBy: 360)
-        return phi > 180 ? 360 - phi : phi
+    public func closestCoordinate(to coordinate: CLLocationCoordinate2D) -> CoordinateAlongPolyline? {
```
1ec5: Let’s turn `CoordinateAlongPolyline` into `Polyline.IndexedCoordinate` or somesuch.

Turf/Turf.swift
```diff
@@ -107,126 +107,117 @@ public struct Turf {
...
-        return coords
+    public static func differenceBetweenAngles(_ alpha: CLLocationDirection, _ beta: CLLocationDirection) -> CLLocationDirection {
+        let phi = abs(beta - alpha).truncatingRemainder(dividingBy: 360)
+        return phi > 180 ? 360 - phi : phi
     }
+}
+
+public struct Polyline {
+    
+    public var coordinates: [CLLocationCoordinate2D]
```
1ec5: This property and the overall struct could use some short documentation comments.

frederoni:
After a second thought, I don't see many advantages of using a polyline struct instead of extending `[CLLocationCoordinate2D]` which would make it shorter and cleaner. Most of the feedback would still be relevant except for the initializer.

1ec5:
Extending `[CLLocationCoordinate2D]`, while arguably more Swifty, could also make the methods less discoverable. Regarding https://github.com/mapbox/turf-swift/issues/2#issuecomment-315217003, it wouldn’t be possible to extend an Objective-C type in the same way, but an MBPolyline class similar to this Polyline struct could be used in that case.

1ec5:
Keeping a distinct Polyline struct would enable us to distinguish between coordinates used as polylines versus polygons versus point collections. These distinctions would become relevant as soon as we port Turf functions meant only for polygons.

frederoni:
I think less discoverability is perfectly fine in a library like this. A simple bullet list can cover the entire API. Functions dedicated to polygons might be a problem though..

Turf/Turf.swift
```diff
@@ -125,8 +118,26 @@ public struct Turf {
     }
 }
 
+/**
+ A `Polyline` struct represents a shape consisting of two or more coordinates,
+ specified as [CLLocationCoordinate2D]
```
1ec5: Nit: place `[CLLocationCoordinate2D]` in backticks.

Turf/Turf.swift
```diff
@@ -125,8 +118,26 @@ public struct Turf {
...
     
+    /**
+     `IndexedCoordinate` is a coordinate with additional information such as
+     the index from its position in the polyline and distance from the start
+     of the polyline.
+     */
+    public struct IndexedCoordinate {
+        let coordinate: Array<CLLocationCoordinate2D>.Element
+        let index: Array<CLLocationCoordinate2D>.Index
+        let distance: CLLocationDistance
```
1ec5: Each of the members of this struct needs minimal documentation.

PR #10, opened 2017-08-11
Add support for macOS

frederoni:
Added support for macOS.
`CLLocationDistanceMax` is not provided in CoreLocation for macOS so I added
```swift
#if os(OSX)
let CLLocationDistanceMax: CLLocationDistance = .greatestFiniteMagnitude
#endif
```

@bsudekum @1ec5 👀


Turf.xcodeproj/project.pbxproj
```diff
@@ -27,6 +38,9 @@
 /* End PBXContainerItemProxy section */
 
 /* Begin PBXFileReference section */
+		353E9B011F3E08BD007CFA23 /* Turf copy-Info.plist */ = {isa = PBXFileReference; lastKnownFileType = text.plist.xml; name = "Turf copy-Info.plist"; path = "/Users/fredrik/Workspace/mapbox/turf-swift/Turf copy-Info.plist"; sourceTree = "<absolute>"; };
```
1ec5: Are you sure it’s OK to use an absolute path here?

I’ve found that the same Info.plist can almost always be shared among iOS, macOS, tvOS, and watchOS targets, so it may be unnecessary to even have a Turf copy-Info.plist around.

frederoni: I overlooked this. Creating a reference or copying while dragging is randomly picked in Xcode 9 beta 1-6. The recommended workaround is to hold down <kbd>⌘</kbd><kbd>⌥</kbd> while dragging for just creating a ref. Anyways, good catch.

PR #14, opened 2017-09-25
Fix access control

frederoni:
Fix access control and reorganize wrap and differenceBetweenAngles.

@1ec5 @bsudekum 👀

### DIFF
--- a/Turf/Turf.swift
+++ b/Turf/Turf.swift
@@ -45,26 +45,16 @@ public struct RadianCoordinate2D {
                                                cos(distance) - sin(latitude) * sin(otherLatitude))
         return RadianCoordinate2D(latitude: otherLatitude, longitude: otherLongitude)
     }
+    
+    /**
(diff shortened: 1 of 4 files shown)


Turf/Turf.swift
```diff
@@ -57,7 +61,7 @@ func -(left: RadianCoordinate2D, right: RadianCoordinate2D) -> RadianDistance {
 /**
  Returns the Haversine distance between two coordinates measured in degrees.
  */
-func -(left: CLLocationCoordinate2D, right: CLLocationCoordinate2D) -> CLLocationDistance {
+public func -(left: CLLocationCoordinate2D, right: CLLocationCoordinate2D) -> CLLocationDistance {
```
1ec5: When I originally wrote this operator, I thought I was being clever, but it turned out to be a mistake. It's too difficult to spot when a distance is being computed, and this operator is symmetric, violating the usual arithmetic definition of subtraction. Let's replace these operators with standard `distance(to:)` methods.

PR #16, opened 2017-10-02
Fix pod lint issues

frederoni:
Fixes pod lint issues and bumps the version

@bsudekum @1ec5 👀


turf-swift.podspec
```diff
@@ -6,8 +6,8 @@ Pod::Spec.new do |s|
   s.version = "0.0.3"
```
1ec5: [The marketing version in Info.plist](https://github.com/mapbox/turf-swift/blob/master/Turf/Info.plist#L18) should match this version.

1ec5:
FYI, publishing this pod to CocoaPods trunk will lock in the `turf-swift` name and module name. That’s OK because it leaves open the possibility of an Objective-C-specific pod in the future.

frederoni:
Considering the wrappers we've talked about in #2, a more generic name would make sense.
Would just changing the pod name in the podspec be sufficient?

1ec5:
> Would just changing the pod name in the podspec be sufficient?

Yep, that’s all that CocoaPods cares about – not even the podspec’s file name matters.

PR #18, opened 2017-10-26
Map Turf.js packages to Turf-swift methods

1ec5:
Added a table to the readme mapping Turf.js packages to Turf-swift methods as best I can.

/cc @frederoni @DenisCarriere


DenisCarriere:
👍

PR #21, opened 2017-11-10
[WIP] Add polygon area calculation

### LINKED ISSUE #20: Polygon Area calculation?  --> support
Did the existing turf has polygon area calculation?

captainbarbosa:
Closes #20 by adding the ability to calculate the area of a polygon.

Turf.js reference: https://github.com/Turfjs/turf/blob/master/packages/turf-area/index.js


Turf/Turf.swift
```diff
@@ -285,3 +285,65 @@ public struct Polyline {
         return closestCoordinate
     }
 }
+
+public struct Polygon {
+    var radius:Float = 6378137
+    var coordinates: [CLLocationCoordinate2D]
+    
+    var polygonArea: Float {
```
frederoni: Let's change all Float to Double so we don't have to cast when converting to radians.

Turf/Turf.swift
```diff
@@ -285,3 +285,65 @@ public struct Polyline {
...
+                    middleIndex = index + 1
+                    upperIndex = index + 2
+                }
+                
+                p1 = coordinates[lowerIndex]
+                p2 = coordinates[middleIndex]
+                p3 = coordinates[upperIndex]
+                area += (p3.longitude.toRadians() - p1.longitude.toRadians()) * sin(p2.latitude.toRadians())
+            }
+            var RADIUS = 6378137
```
frederoni: nit: radius is already defined in the Polygon struct.

Turf/Turf.swift
frederoni: After a second thought, this should probably be `CLLocationDistance`

Turf/Turf.swift
```diff
@@ -285,3 +285,65 @@ public struct Polyline {
         return closestCoordinate
     }
 }
+
+public struct Polygon {
+    var radius:Float = 6378137
```
frederoni: Let's rename this to `equatorialRadius` for clarity and move it to the top where `metersPerRadian` is.

frederoni:
We should probably migrate this project to an Xcode 9 stack on Bitrise.

Turf/Turf.swift
```diff
@@ -285,3 +286,61 @@ public struct Polyline {
         return closestCoordinate
     }
 }
+
+public struct Polygon {
+    var coordinates: [CLLocationCoordinate2D]
```
frederoni: Capturing from chat: The coordinates property on a GeoJSON Polygon is represented as a `[[CLLocationCoordate2D]]`

Turf/Turf.swift
```diff
@@ -285,3 +288,67 @@ public struct Polyline {
...
+    
+    var polygonArea: Double {
+        var area:Double = 0
+   
+        if (!polygonCoordinates.isEmpty && polygonCoordinates.count > 0) {
+            
+            area += abs(ringArea(polygonCoordinates[0]))
+
+            for coordinate in polygonCoordinates.suffix(from: 1) {
+                area -= abs(ringArea(coordinate))
```
1ec5: Note that each item in `polygonCoordinates` represents a ring’s coordinates, not an individual `CLLocationCoordinate2D`.

A more functional approach would be:

```swift
let area = abs(ringArea(polygonCoordinates.first ?? []))
    - polygonCoordinates.suffix(from: 1)
        .map { abs(ringArea($0)) } // convert the inner rings to their areas
        .reduce(0, +) // sum the inner areas
```

Turf/Turf.swift
```diff
@@ -285,3 +288,67 @@ public struct Polyline {
...
+        var p1: CLLocationCoordinate2D
+        var p2: CLLocationCoordinate2D
+        var p3: CLLocationCoordinate2D
+        var lowerIndex: Int
+        var middleIndex: Int
+        var upperIndex: Int
+        var area: Double = 0
+        let coordinatesCount: Int = coordinates.count
+        
+        if (coordinatesCount > 2) {
```
1ec5: Nit: drop the parentheses (×3).

Turf/Turf.swift
```diff
@@ -285,3 +288,67 @@ public struct Polyline {
         return closestCoordinate
     }
 }
+
+public struct Polygon {
+    var polygonCoordinates: [[CLLocationCoordinate2D]]
```
1ec5: GeoJSON represents a polygon as a collection of rings (one outer ring and multiple inner rings). Similarly, the Mapbox Maps SDKs for iOS and macOS represent a polygon as an outer polygon containing multiple inner polygons. Either way, it’s nice to be able to work with a formal structure instead of a nested array. Consider replacing this property with:

```swift
var rings: [Ring]
```

or:

```swift
var outerRing: Ring
var innerRings: [Ring]
```

where `Ring` is either a struct or a type alias for `[CLLocationCoordinate2D]`. As a struct, it would be a good home for the `ringArea(_:)` method.

Turf/Turf.swift
```diff
@@ -285,3 +288,67 @@ public struct Polyline {
...
+    
+    var polygonArea: Double {
```
1ec5: This property already lives inside a struct named `Polygon`, so the word “polygon” here is [redundant](https://swift.org/documentation/api-design-guidelines/#omit-needless-words).

Turf/Turf.swift
```diff
@@ -285,3 +288,67 @@ public struct Polyline {
...
+        var p2: CLLocationCoordinate2D
+        var p3: CLLocationCoordinate2D
+        var lowerIndex: Int
+        var middleIndex: Int
+        var upperIndex: Int
+        var area: Double = 0
+        let coordinatesCount: Int = coordinates.count
+        
+        if (coordinatesCount > 2) {
+            for index in 0...coordinatesCount - 1 {
```
1ec5: `0..<coordinatesCount` would be a little clearer.

Turf/Turf.swift
1ec5: Add a comment with a permalink to the specific commit of the Turf package you’re porting this code from, in case we need to track down a discrepancy in the future.

Turf/Turf.swift
```diff
@@ -285,3 +288,67 @@ public struct Polyline {
...
+     * Laboratory, Pasadena, CA, June 2007 http://trs-new.jpl.nasa.gov/dspace/handle/2014/40409
+     *
+     */
+    private func ringArea(_ coordinates: [CLLocationCoordinate2D]) -> Double {
+        var p1: CLLocationCoordinate2D
+        var p2: CLLocationCoordinate2D
+        var p3: CLLocationCoordinate2D
+        var lowerIndex: Int
+        var middleIndex: Int
+        var upperIndex: Int
```
1ec5: These six variables are always written inside each iteration of the `for` loop and read immediately thereafter, and never read outside the loop. Move these variables to inside the loop. The original turf-area implementation declares these variables upfront to avoid [hoisting](https://developer.mozilla.org/en-US/docs/Glossary/Hoisting), but it isn’t even necessary to do so in modern, strict-mode JavaScript.

1ec5: Use a tuple consisting of three coordinates to ensure that all three are set simultaneously:

```swift
let controlPoints: (CLLocationCoordinate2D, CLLocationCoordinate2D, CLLocationCoordinate2D)
if index == coordinatesCount - 2 {
    controlPoints = (coordinates[coordinatesCount - 2], coordinates[coordinatesCount - 1], coordinates[0])
}
// …
```

Turf/Turf.swift
```diff
@@ -285,3 +288,67 @@ public struct Polyline {
...
+                    upperIndex = index + 2
+                }
+                
+                p1 = coordinates[lowerIndex]
+                p2 = coordinates[middleIndex]
+                p3 = coordinates[upperIndex]
+                area += (p3.longitude.toRadians() - p1.longitude.toRadians()) * sin(p2.latitude.toRadians())
+            }
+            
+            area = area * Constants.equatorialRadius * Constants.equatorialRadius / 2
```
1ec5: Nit: use the `*=` operator.

Turf/Turf.swift
```diff
@@ -285,3 +288,67 @@ public struct Polyline {
...
     }
 }
+
+public struct Polygon {
+    var polygonCoordinates: [[CLLocationCoordinate2D]]
+    
+    var polygonArea: Double {
+        var area:Double = 0
+   
+        if (!polygonCoordinates.isEmpty && polygonCoordinates.count > 0) {
```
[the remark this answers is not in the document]
1ec5: If we use a non-optional `outerRing` properties as in https://github.com/mapbox/turf-swift/pull/21#discussion_r155851566, then we can assume there’s an outer ring and won’t need the `if` statement at all.

Turf/Turf.swift
1ec5: If we use separate `outerRing` and `innerRings` properties as in https://github.com/mapbox/turf-swift/pull/21#discussion_r155851566, then this simplifies further to:

```swift
let area = abs(ringArea(outerRing))
    - innerRings
        .map { abs(ringArea($0)) } // convert the inner rings to their areas
        .reduce(0, +) // sum the inner areas
```

Turf/Turf.swift
```diff
@@ -315,40 +300,47 @@ public struct Polygon {
...
      *
      */
-    private func ringArea(_ coordinates: [CLLocationCoordinate2D]) -> Double {
-        var p1: CLLocationCoordinate2D
-        var p2: CLLocationCoordinate2D
-        var p3: CLLocationCoordinate2D
-        var lowerIndex: Int
-        var middleIndex: Int
-        var upperIndex: Int
+    internal func area() -> Double {
```
1ec5: Turn this method into a computed property to reduce the friction of using it.

Turf/Turf.swift
```diff
@@ -4,10 +4,9 @@ public typealias LocationRadians = Double
 public typealias RadianDistance = Double
 public typealias RadianDirection = Double
 
-struct Constants {
-    static let metersPerRadian = 6_373_000.0
-    static let equatorialRadius:Double = 6378137
-}
+
+let metersPerRadian: CLLocationDistance = 6_373_000.0
+let equatorialRadius: CLLocationDistance = 6378137
```
1ec5: Nit: similar to in `metersPerRadian`, use underscores to group the digits.

1ec5: `equatorialRadius` is currently set to the equatorial radius according to WGS84 and the [IUGG](https://github.com/Project-OSRM/osrm-backend/blob/088d4edc6b761d2642a9b1ab7d85b701d3165fb2/include/util/coordinate_calculation.hpp#L26-L28). This value is used in [OSRM](https://github.com/Project-OSRM/osrm-backend/blob/1d4d3b80b5e53fb342a73b5556c9dd3e78d5c4c7/include/util/web_mercator.hpp#L19) and [mbgl](https://github.com/mapbox/mapbox-gl-native/blob/0fddad6e0b828af2d49d65953d31caaad8d8c953/include/mbgl/util/constants.hpp#L32) for Web Mercator calculations.

Since different radii are used in different formulas and the errors can sometimes build up, let’s document which is which.

Turf/Turf.swift
```diff
@@ -4,10 +4,9 @@ public typealias LocationRadians = Double
 public typealias RadianDistance = Double
 public typealias RadianDirection = Double
 
-struct Constants {
-    static let metersPerRadian = 6_373_000.0
-    static let equatorialRadius:Double = 6378137
-}
+
+let metersPerRadian: CLLocationDistance = 6_373_000.0
```
1ec5: Apparently this value of 6&nbsp;373&nbsp;000&nbsp;m came from [an old version of turf-distance](https://github.com/Turfjs/turf-distance/blob/cea4c6cbec582bff8880193a3c2cda38ae731967/index.js#L78) that I originally ported to an internal Swift application, before it got copied into another internal Swift application, then copied into the navigation SDK, then moved here. It’s close to the value of 6&nbsp;372&nbsp;797.560&nbsp;856 that [Osmium](https://github.com/Project-OSRM/osrm-backend/blob/bb1f4a025a3cefd3598a38b9d3e55485d1080ec5/third_party/libosmium/include/osmium/geom/haversine.hpp#L58-L59) describes as “Earth’s quadratic mean radius for WGS84”.

These days, Turf.js uses [a spherical approximation](https://github.com/Turfjs/turf/blob/a94151418cb969868fdb42955a19a133512da0fd/packages/turf-helpers/index.js#L4) of 6&nbsp;671&nbsp;008.8&nbsp;m for its radian-to-meter conversion and Haversine formula. The Haversine formula was always meant to be used with a spherical meters-per-radian value. We should probably change this value to match Turf.js. 🙀

/ref Turfjs/turf#978 Turfjs/turf#1012 Turfjs/turf#1176
/cc @frederoni @bsudekum

TurfTests/TurfTests.swift
```diff
@@ -291,12 +291,13 @@ class TurfTests: XCTestCase {
...
         let geometry = json["geometry"] as! [String: Any]
         let geoJSONCoordinates = geometry["coordinates"] as! [[[Double]]]
-        let coordinates = geoJSONCoordinates.map {
+        let allRings = geoJSONCoordinates.map {
            $0.map { CLLocationCoordinate2D(latitude: $0[1], longitude: $0[0]) }
         }
+        let outerRing = Ring(coordinates: allRings.first!)
         
-        let polygon = Polygon(polygonCoordinates: coordinates)
+        let polygon = Polygon(outerRing: outerRing, innerRings: [])
```
1ec5: What about the inner rings in `allRings`?

Turf/Turf.swift
1ec5: Since the polygon area calculation doesn’t depend on `metersPerRadian`, we can treat this change as tail work: #26.

captainbarbosa: If we're going to change `metersPerRadian ` to `6 671 008.8 m`, should this happen in a different PR since it would globally impact this library? Or are you ok with me making the change here?

1ec5: Yeah, let’s do it in a separate PR for #26, so that any fallout is easier to track down.

In the meantime, can you add some comments explaining why these two constants differ?

TurfTests/TurfTests.swift
```diff
@@ -295,9 +295,10 @@ class TurfTests: XCTestCase {
            $0.map { CLLocationCoordinate2D(latitude: $0[1], longitude: $0[0]) }
         }
         let outerRing = Ring(coordinates: allRings.first!)
+        let innerRing = Ring(coordinates: allRings[1])
```
1ec5: `innerRings` can be expressed as `allRings.prefix(from: 1)` – handy for dealing with a two-hole polygon.

captainbarbosa: It looks like `prefix(from:)` isn't available in Swift 4.

captainbarbosa: Ooooo I think you meant `suffix(from: 1)` 😉

1ec5: Yeah that gets me every time. 😄

1ec5:
91907e37ca7d1e998a6d5669eaa3d477d2e72100 adds an entry for `Polygon.area` to the readme.

PR #24, opened 2017-11-27
Upgrade to Swift 4

bsudekum:
Since for the most part everything is contained in a struct, there was no need to add `@objc` since structs do not bridge to objc.

/cc @frederoni @friedbunny


Turf.xcodeproj/project.pbxproj
```diff
@@ -240,7 +240,7 @@
 			isa = PBXProject;
 			attributes = {
 				LastSwiftUpdateCheck = 0900;
-				LastUpgradeCheck = 0900;
+				LastUpgradeCheck = 0910;
 				ORGANIZATIONNAME = "Fredrik Karlsson";
```
friedbunny: @frederoni 🤔 😝

PR #31, opened 2018-01-17
Rewrite installation instructions

1ec5:
Rewrote the “Installation” section of the readme, adding sections for Carthage and SPM. Also, the name of the pod is `Turf`, not `Turf-swift`.

/cc @frederoni @captainbarbosa

PR #32, opened 2018-01-18
Support macOS, tvOS, watchOS

### LINKED ISSUE #8: macOS, tvOS, watchOS
This library only supports iOS, but there's no technical reason it couldn't also support macOS, tvOS, and watchOS. The project and podspec would both need to be extended for these platforms. See MapboxGeocoder.swift, MapboxDirections.swift, and MapboxStatic.swift for examples.

/cc @frederoni @bsudekum @boundsj @friedbunny

1ec5:
This PR expands this library’s platform support to include deploying on:

* iOS 8 when installing via Carthage
* macOS 10.10 and above when installing via Carthage
* tvOS 11 and above
* watchOS 2 and above

Along the way, the Mac targets have been upgraded to Swift 4, and all the bundle identifiers are now consistent.

Fixes #8.

/cc @frederoni @friedbunny @captainbarbosa

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 📱[![](https://www.bitrise.io/app/49f5bcca71bf6c8d/status.svg?token=SzGBTkEtxsbuAnbcF9MTog&branch=master)](https://www.bitrise.io/app/49f5bcca71bf6c8d)
 🖥💻[![](https://www.bitrise.io/app/b72273651db53613/status.svg?token=ODv2UnyAHoOxV8APATEBFw&branch=master)](https://www.bitrise.io/app/b72273651db53613)
 
-A [spatial analysis](http://en.wikipedia.org/wiki/Spatial_analysis) library written in Swift for native iOS, macOS, and tvOS applications, ported from [Turf.js](https://github.com/Turfjs/turf/).
+A [spatial analysis](http://en.wikipedia.org/wiki/Spatial_analysis) library written in Swift for native iOS, macOS, tvOS, and watchOS applications, ported from [Turf.js](https://github.com/Turfjs/turf/).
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -2,8 +2,10 @@
 
 📱[![](https://www.bitrise.io/app/49f5bcca71bf6c8d/status.svg?token=SzGBTkEtxsbuAnbcF9MTog&branch=master)](https://www.bitrise.io/app/49f5bcca71bf6c8d)
 🖥💻[![](https://www.bitrise.io/app/b72273651db53613/status.svg?token=ODv2UnyAHoOxV8APATEBFw&branch=master)](https://www.bitrise.io/app/b72273651db53613)
+📺[![](https://www.bitrise.io/app/0b037542c2395ffb/status.svg?token=yOtMqbu-5bj8grB1Jmoefg)](https://www.bitrise.io/app/0b037542c2395ffb)
+⌚️[![](https://www.bitrise.io/app/0d4d611f02295183/status.svg?token=NiLB_E_0IvYYqV4Mj973TQ)](https://www.bitrise.io/app/0d4d611f02295183)
 
-A [spatial analysis](http://en.wikipedia.org/wiki/Spatial_analysis) library written in Swift for native iOS and macOS applications, ported from [Turf.js](https://github.com/Turfjs/turf/).
+A [spatial analysis](http://en.wikipedia.org/wiki/Spatial_analysis) library written in Swift for native iOS, macOS, tvOS, and watchOS applications, ported from [Turf.js](https://github.com/Turfjs/turf/).
```
frederoni: 🐧

1ec5: Hehe, let’s take care of that in #31, which I’ll also update to reflect these changes.

PR #34, opened 2018-01-18
Yosemite, not Cheetah

1ec5:
Deploying backwards to Mac OS X 10.0 is probably not a great use of our time.

/cc @friedbunny


1ec5:
I’ll consider that a request for turf-applescript. 👍

PR #45, opened 2018-01-29
Codable GeoJSON

### LINKED ISSUE #43: Encode and decode GeoJSON
It’d be [pretty straightforward](https://medium.com/@mrlegowatch/omg-codable-is-so-frickin-awesome-bb9ff33139da) to add general-purpose GeoJSON encoding and decoding functionality to this library using Codable. We could then add conveniences for creating instances of Polyline and Polygon from GeoJSON.

/cc @frederoni @captainbarbosa

frederoni:
Fixes #43 - Encode and decode GeoJSON

TODO:
- Decode, and test
    - [x] PointFeature
    - [x] LineStringFeature
    - [x] PolygonFeature
    - [x] MultiPointFeature
    - [x] MultiLineStringFeature
    - [x] MultiPolygonFeature
    - [x] FeatureCollection
- [x] Encode and verify that all of the above structs round-trips
- [x] Replace Polyline and Polygon with corresponding GeoJSON types
- [x] Fold Polyline into LineString 
- [x] Parse and decode GeoJSON of unknown types

```swift
// Decode unknown GeoJSON type
let geojson = try! GeoJSON.parse(data)

// Decode known GeoJSON type
let geojson = try! GeoJSON.parse(FeatureCollection.self, from: data)
```

@1ec5 @captainbarbosa 👀


Sources/Turf/CoreLocation.swift
```diff
@@ -50,6 +50,20 @@ extension CLLocationDirection {
     }
 }
 
+extension CLLocationCoordinate2D: Codable {
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.unkeyedContainer()
+        try container.encode(longitude)
+        try container.encode(latitude)
```
1ec5: The GeoJSON specification [allows](https://tools.ietf.org/html/rfc7946#section-3.1.1) for a third item in the array, representing the altitude or elevation of the position. As far as I know, Turf.js doesn’t make use of altitude values, but [Mapbox Java Services does support deserializing them](https://github.com/mapbox/mapbox-java/blob/c15c234837c1379d4c7044e9b4c769e0f3d10b39/services-geojson/src/main/java/com/mapbox/geojson/Point.java#L154).

I think this code will gracefully ignore any altitude present in the original GeoJSON, but it won’t round-trip the value. I think that’s fine for now, but FYI an alternative would be to decode the position as a CLLocation, which can store an altitude.

1ec5: When you create a CLLocation with `CLLocation(latitude:longitude:)`, `timestamp` is set to the current time. I suppose that would be problematic.

We could roll our own type instead of piggybacking on Core Location, but CLLocationCoordinate2D and CLLocation come out of the box with nice debugging tools and developer familiarity.

Let’s stick with CLLocationCoordinate2D until we implement a function for which the altitude becomes relevant. I think it’s fine if the GeoJSON serialization comes with a few caveats, since this library isn’t a general-purpose GeoJSON converter.

frederoni:
@1ec5 this PR is ready for a cursory review

Sources/Turf/GeoJSON.swift
```diff
@@ -0,0 +1,358 @@
...
+
+public struct Geometry: Codable {
+    public var type: String
+}
+
+/**
+ A `LineString` struct represents a shape consisting of two or more coordinates,
+ specified as `[CLLocationCoordinate2D]`
+ */
+public struct LineString: Codable {
```
1ec5: Maybe typealias Polyline to LineString (and MultiPolyline to MultiLineString), both for backwards compatibility and to accommodate muscle memory, especially since the map SDK still calls it MGLPolyline.

Sources/Turf/GeoJSON.swift
```diff
@@ -0,0 +1,358 @@
+import Foundation
+#if !os(Linux)
+import CoreLocation
+#endif
+
+
+public class BaseFeature: Codable {
```
1ec5: Purely from the perspective of RFC&nbsp;7946, this would be a protocol adopted by both Feature and FeatureCollection, since the real base class of Geometry, Feature, and FeatureCollection would be GeoJSONObject.

Sources/Turf/GeoJSON.swift
```diff
@@ -0,0 +1,358 @@
...
+        properties = try container.decode([String : AnyJSONType]?.self, forKey: .properties)
+    }
+    
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        try container.encode(properties, forKey: .properties)
+    }
+}
+
+public class Feature: BaseFeature {
```
1ec5: It seems problematic that features and feature collections would have class (by reference) semantics while geometries would have struct (by value) semantics.

Sources/Turf/GeoJSON.swift
```diff
@@ -0,0 +1,358 @@
...
+        } else {
+            throw GeoJSONError.unknownType
+        }
+    }
+    
+    public static func parse(data: Data) throws -> GeoJSON {
+        return try JSONDecoder().decode(GeoJSON.self, from: data)
+    }
+    
+    public static func parse<T: BaseFeature>(data: Data, as: T.Type) throws -> T {
```
1ec5: What if we change this around so that you call e.g. `FeatureCollection(geoJSONData:)` if you expect it to be a feature collection?

Sources/Turf/GeoJSON.swift
```diff
@@ -0,0 +1,358 @@
...
+    
+    init(coordinates: [[CLLocationCoordinate2D]]) {
+        self.coordinates = coordinates
+    }
+    
+    var innerRings: [Ring]? {
+        get { return Array(coordinates.suffix(from: 1)).map { Ring(coordinates: $0) } }
+    }
+    
+    var outerRing: Ring? {
```
1ec5: A polygon without an outer ring is degenerate. Is there anything we can do to make this property non-optional?

Sources/Turf/GeoJSON.swift
```diff
@@ -0,0 +1,358 @@
...
+
+public struct Polygon: Codable {
+    var type: String = GeometryType.Polygon.rawValue
+    var coordinates: [[CLLocationCoordinate2D]]
+    
+    init(coordinates: [[CLLocationCoordinate2D]]) {
+        self.coordinates = coordinates
+    }
+    
+    var innerRings: [Ring]? {
```
1ec5: What’s the difference between a `innerRings` being `nil` and `innerRings` being an empty array?

Sources/Turf/GeoJSON.swift
```diff
@@ -0,0 +1,358 @@
...
+}
+
+public struct MultiLineString: Codable {
+    var type: String = GeometryType.MultiLineString.rawValue
+    var coordinates: [[CLLocationCoordinate2D]]
+}
+
+public struct MultiPolygon: Codable {
+    var type: String = GeometryType.MultiLineString.rawValue
+    var coordinates: [[[CLLocationCoordinate2D]]]
```
1ec5: `[[Ring]]` would give us more flexibility and be less error-prone than `[[[CLLocationCoordinate2D]]]`. It’s a bit unfortunate that we can’t just have this be `[Polygon]`, due to the absence of `type` properties on the items of the top-level array.

Sources/Turf/GeoJSON.swift
```diff
@@ -0,0 +1,358 @@
...
+    }
+    
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        try container.encode(properties, forKey: .properties)
+    }
+}
+
+public class Feature: BaseFeature {
+    public var type: GeoJSONType
```
1ec5: One of the caveats we’re introducing is that, if you deserialize and serialize again, all the feature IDs are lost. IDs are quite common in GeoJSON, so losing them would be a surprise. Consider adding an `identifier` property to hold the ID.

Sources/Turf/GeoJSON.swift
```diff
@@ -0,0 +1,358 @@
...
+        
+        if feature.type == .Feature {
+            guard let type = GeometryType(rawValue: feature.simplifiedGeometry!.type) else {
+                throw GeoJSONError.unknownType
+            }
+            self.value = try container.decode(type.featureType)
+        } else if feature.type == .FeatureCollection {
+            self.value = try container.decode(FeatureCollection.self)
+        } else {
+            throw GeoJSONError.noTypeFound
```
1ec5: RFC&nbsp;7946 allows a “GeoJSON text” (a data blob) to consist of a bare Geometry, even a bare Position. Does this GeoJSON class support deserializing a Polygon or `CLLocationCoordinate2D` at the top level?

Sources/Turf/GeoJSON.swift
frederoni: Polyline to LineString compatibility alias should be sufficient since MultiPolyline is being introduced in this PR.

1ec5: OK, the map SDK has MGLMultiPolyline too, but I suppose that isn’t as well-known as MGLPolyline.

Sources/Turf/GeoJSON.swift
```diff
@@ -3,56 +3,96 @@ import Foundation
 import CoreLocation
 #endif
 
+public typealias FeatureIdentifierNumber = Numeric & Codable
+public typealias FeatureIdentifierString = StringProtocol & Codable
 
-public class BaseFeature: Codable {
-    private enum CodingKeys: String, CodingKey {
-        case properties
+public struct FeatureIdentifier<T: FeatureIdentifierNumber, U: FeatureIdentifierString>: Codable {
```
1ec5: Would it be possible to represent the identifier as an enumeration with associated values (with cases `number(Numeric)` and `string(StringProtocol)`)? I think that would be easier for a developer to understand than a struct with template parameters. Or would that be incompatible with Codable?

Sources/Turf/GeoJSON.swift
1ec5: For context, the main reason we originally named the class MGLPolyline instead of MGLLineString was for consistency with MapKit, but LineString also sounded too similar to String, especially in the context of an application that uses Core Text.

Sources/Turf/GeoJSON.swift
```diff
@@ -0,0 +1,411 @@
+import Foundation
+#if !os(Linux)
+import CoreLocation
+#endif
+
+public typealias FeatureIdentifierNumber = Numeric & Codable
+public typealias FeatureIdentifierString = StringProtocol & Codable
+
+public struct FeatureIdentifier<T: FeatureIdentifierNumber, U: FeatureIdentifierString>: Codable {
```
1ec5: Would it be possible to represent the identifier as an enumeration with associated values (with cases `number(Numeric)` and `string(StringProtocol)`)? I think that would be easier for a developer to understand than a struct with template parameters. Or would that be incompatible with Codable?

Sources/Turf/GeoJSON.swift
```diff
@@ -0,0 +1,411 @@
...
+                        featureTypes.append(MultiPolygonFeature.self)
+                    case .MultiPoint:
+                        featureTypes.append(MultiPointFeature.self)
+                    }
+                }
+            }
+        }
+        
+        if var mappedContainer = try? container.nestedUnkeyedContainer(forKey: .features) {
+            while (!mappedContainer.isAtEnd) {
```
1ec5: Nit: unnecessary parentheses.

Sources/Turf/GeoJSON.swift
```diff
@@ -0,0 +1,411 @@
...
+                } else if featureType is PolygonFeature.Type {
+                    features.append(try mappedContainer.decode(PolygonFeature.self))
+                } else if featureType is PointFeature.Type {
+                    features.append(try mappedContainer.decode(PointFeature.self))
+                } else if featureType is MultiPoint.Type {
+                    features.append(try mappedContainer.decode(MultiPointFeature.self))
+                } else if featureType is MultiLineString.Type {
+                    features.append(try mappedContainer.decode(MultiLineStringFeature.self))
+                } else if featureType is MultiPolygon.Type {
+                    features.append(try mappedContainer.decode(MultiPolygonFeature.self))
```
1ec5: This is pretty repetitive. Could we use `type(of:)` to consolidate all these cases into one case?

Sources/Turf/GeoJSON.swift
frederoni: Protocols with associated values are incompatible with Codable but enumerations with associated values should work. However, we'd lose some type safety due to the lack of support for raw values but it would still be an improvement.

Sources/Turf/GeoJSON.swift
frederoni: Could help with discoverability but it would bloat the code a bit because it would mean one implementation per feature/geometry. I figured a similar signature to JSONDecoder’s decode would make more sense so I renamed this to `GeoJSON.parse(_:from:)`

Sources/Turf/GeoJSON.swift
```diff
@@ -3,49 +3,47 @@ import Foundation
...
-
-public struct FeatureIdentifier<T: FeatureIdentifierNumber, U: FeatureIdentifierString>: Codable {
-    var t: T?
-    var u: U?
-    
-    public var value: Codable {
-        return t ?? u
+public enum FeatureIdentifier {
+    case string(String)
+    case int(Int)
```
1ec5: An identifier can be any number, not necessarily an integer.

Sources/Turf/GeoJSON.swift
frederoni: Fixed in https://github.com/mapbox/turf-swift/pull/45/commits/c3edfbee5d0c2ab0c731009f2f9f518c2e0d8e53

PR #52, opened 2018-06-01
Bearing

RomainQuidet:
add support of bearing computation


frederoni:
Thanks for the PR. It looks like we already provide bearing/direction between two coordinates in an extension on `CLLocationCoordinate2D.direction(to:)`.

https://github.com/mapbox/turf-swift/blob/4aebfdb14cef5e3512fc9a2a550ebba4c2edf4d6/Sources/Turf/CoreLocation.swift#L79-L82

RomainQuidet:
what a fool, I searched in Turf, not in CoreLocation extensions ...

PR #53, opened 2018-06-07
Add support for xcode 10, only import swift files

bsudekum:
If this is left wide open, all and any file type get copied over when using cocoapods. If used in conjunction with xcode 10, it fails to compile and complains it's copied twice.

Should this also include .h files?

/cc @1ec5 @frederoni


frederoni:
> Should this also include .h files?

Yep, we should add the `Turf.h` including future header files.

PR #54, opened 2018-06-20
Refactor

### LINKED ISSUE #42: Split up Turf.swift
Originally, this library was contained in a single file to facilitate manual installation without a package manager. However, as this library grows, it’ll become unwieldy to work with this file. We should move each struct into a file named after the struct. As for the `Turf` struct, we should replace `intersection(_:_:)` with a method on `Polyline` that takes a single other `Polyline` to intersect with.

frederoni:
Fixes #42 Split up into separate files including tests

- [ ] replace `intersection(_:_:)` with a method on Polyline that takes a single other Polyline to intersect with

👆 is not a simple rename. Should we do that as tail work so that it's more reviewable?

cc @1ec5


bsudekum:
Cool if we put out a patch release with https://github.com/mapbox/turf-swift/pull/53 before merging this? Can queue up now.

frederoni:
@bsudekum sure

PR #65, opened 2018-09-18
Fixed erroneous initializer in the previous commit

SandyChapman:
Not sure how I missed this, but the default values in the reduce func in the initializer causes a lot of problems for things not bounding around the lat lon of (0, 0).


frederoni:
Looks like you have to re-sync your fork and rebase this PR.

Let's make `BoundingBox.northWest` and `BoundingBox.southEast` public.

SandyChapman:
New PR on #66.

PR #66, opened 2018-09-19
Fixed erroneous initializer in the previous commit

SandyChapman:
Rebased https://github.com/mapbox/turf-swift/pull/65

Not sure how I missed this, but the default values in the reduce func in the initializer causes a lot of problems for things not bounding around the lat lon of (0, 0).

Made BoundingBox.northWest and BoundingBox.southEast public and mutable.

### DIFF
--- a/Sources/Turf/BoundingBox.swift
+++ b/Sources/Turf/BoundingBox.swift
@@ -36,6 +36,6 @@ public struct BoundingBox: Codable {
     
     // MARK: - Private
     
-    let northWest: CLLocationCoordinate2D
-    let southEast: CLLocationCoordinate2D
+    public var northWest: CLLocationCoordinate2D
+    public var southEast: CLLocationCoordinate2D
 }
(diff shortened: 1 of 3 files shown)


Sources/Turf/BoundingBox.swift
```diff
@@ -9,8 +9,9 @@ public struct BoundingBox: Codable {
         guard coordinates?.count ?? 0 > 0 else {
             return nil
         }
+        let startValue = (minLat: coordinates!.first!.latitude, maxLat: coordinates!.first!.latitude, minLon: coordinates!.first!.longitude, maxLon: coordinates!.first!.longitude)
```
frederoni: Abbreviations should be avoided according to [Swift’s API design guidelines](https://swift.org/documentation/api-design-guidelines/) but min/max and lat/lon is a common standard and this implementation is private so it's fine.

SandyChapman: @frederoni : Agreed on the abbrevs. I generally follow the guideline that variable names are proportional to the size of the scope they're used in. I.e. small scope, all visible on screen means a variable name can be abbreviated if following common practice.

PR #67, opened 2018-09-21
Indexed coordinate from start of line string

1ec5:
Supplemented `LineString.coordinateFromStart(distance:)` with an implementation that returns an `IndexedCoordinate`, making it easier for clients to stop assuming that a line string never doubles back on itself.

/cc @frederoni

### DIFF
--- a/Sources/Turf/LineString.swift
+++ b/Sources/Turf/LineString.swift
@@ -110,11 +110,21 @@ extension LineString {
      Returns a coordinate along a LineString at a certain distance from the start of the polyline.
      */
     public func coordinateFromStart(distance: CLLocationDistance) -> CLLocationCoordinate2D? {
+        return indexedCoordinateFromStart(distance: distance)?.coordinate
+    }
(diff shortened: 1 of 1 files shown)


Sources/Turf/LineString.swift
```diff
@@ -125,17 +135,18 @@ extension LineString {
...
-                return coordinates[i].coordinate(at: overshoot, facing: direction)
+                let coordinate = coordinates[i].coordinate(at: overshoot, facing: direction)
+                return IndexedCoordinate(coordinate: coordinate, index: i - 1, distance: distance)
             }
             
             traveled += coordinates[i].distance(to: coordinates[i + 1])
         }
         
-        return coordinates.last
+        return IndexedCoordinate(coordinate: coordinates.last!, index: coordinates.endIndex - 1, distance: traveled)
```
frederoni: shouldn't this just be `coordinates.endIndex`?

PR #69, opened 2018-09-25
Added GeoJSON feature initializers

frederoni:
Added public initializers for `LineString`,`MultiLineString`, `MultiLineStringFeature`, and `FeatureCollection` including some tests.

@1ec5

### DIFF
--- a/Sources/Turf/MultiLineString.swift
+++ b/Sources/Turf/MultiLineString.swift
@@ -10,6 +10,10 @@ import CoreLocation
 public struct MultiLineString: Codable, Equatable {
     var type: String = GeometryType.MultiLineString.rawValue
     public var coordinates: [[CLLocationCoordinate2D]]
+    
+    public init(_ coordinates: [[CLLocationCoordinate2D]]) {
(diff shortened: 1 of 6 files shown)


Sources/Turf/MultiLineString.swift
```diff
@@ -10,6 +10,10 @@ import CoreLocation
 public struct MultiLineString: Codable, Equatable {
     var type: String = GeometryType.MultiLineString.rawValue
     public var coordinates: [[CLLocationCoordinate2D]]
+    
+    public init(_ coordinates: [[CLLocationCoordinate2D]]) {
```
1ec5: Should there be a more object-oriented representation like `[LineString]`? Not a dealbreaker for this PR, but it seems less than ideal that we can’t better prevent against developers accidentally passing in malformed coordinate arrays that aren’t line strings.

1ec5: Yeah, I was only thinking of a convenience initializer that maps `LineString`s to their `coordinates`es.

Issue #27, opened 2018-01-02
Inconsistent results from Polyline.distance


systemed:
I'm getting small inconsistencies from Polyline.distance. By tracing intermediate points along a polyline from start to end - including between vertices - the resulting `distance(to: point)` doesn't consistently increment from 0 to (total distance), but jumps up and down.

My suspicion is that there's something screwy going on with the interaction with closestCoordinate but I've not been able to nail it down.

For my purposes I only need the distance to a point from the start of the polyline, so I was able to get round it by hacking a `baseIndex` property onto closestCoordinate which always contains the value of `index` (not `index+1`) and then building this round it:

````swift
let slice = polyline.coordinates[0 ... coord.baseIndex]
let zipped = zip(slice.prefix(upTo: slice.count - 1), slice.suffix(from: 1))
return zipped.map { $0.distance(to: $1) }.reduce(0, +) +
	polyline.coordinates[coord.baseIndex].distance(to: coord.coordinate)
````

but I thought it worth flagging in case anyone else encounters this.

1ec5:
I wonder if this is related to the Earth radius constant we’re using: #26.

/cc @frederoni @bsudekum

KaterGreg:
I'm having the same issue in 0.2.0.

Calculating distance from 2 different points less than 1 meter apart are giving me results that are 90 meters apart.

Original PolyLine:
49.119869999999999, -122.65477
49.119949999999996, -122.65477
49.120129999999996, -122.65477
49.120129999999996, -122.65401
49.12032, -122.65401
49.120689999999996, -122.65401
49.120619999999995, -122.65352 - 183.04443165794757
49.120189999999994, -122.65237 - 279.46026229062824
49.120149999999995, -122.65227 - 287.99193938810771
49.119779999999999, -122.65275 - 341.98021101023426
49.119639999999997, -122.65293 - 362.33204275778985
— more data points follow, but are irrelevant for this example —

Location = **49.120403526377203, -122.6529443631224**
polyline.distance(from: polyline.coordinates.first, to: location.coordinate) = **327.54650624173649**

polylineIndex = polyline.closestCoordinate(to: location.coordinate)!
polylineIndex.coordinate = **49.120404457012796, -122.65294354782493** 
polyline.distance(from: polyline.coordinates.first, to: polylineIndex.coordinate) = **327.54632204561312**

Fake Point (calculated mid-point between points 7 & 8) = **49.120405, -122.652945**
[...]

PR #73, opened 2018-12-10
Bezier spline

RomainQuidet:
Add bezier spline support by porting original code


Sources/Turf/Turf.swift
```diff
@@ -41,4 +41,25 @@ public struct Turf {
         let intersectsWithLine2 = b > 0 && b < 1
         return intersectsWithLine1 && intersectsWithLine2 ? intersection : nil
     }
+    
+    /**
+     Returns a new LineString based on bezier transformation of the input line
+     */
+    public static func bezier(_ line: LineString, resolution: Int = 10000, sharpness: Double = 0.85) -> LineString? {
```
frederoni: This static function would probably be a better fit as a method `LineString.bezier(resolution:sharpness)` in `LineString.swift`.

Sources/Turf/Spline.swift
```diff
@@ -0,0 +1,127 @@
+import Foundation
+#if !os(Linux)
+import CoreLocation
+#endif
+
+
+/**
```
frederoni: nit: Add documentation or remove the block if `SplinePoint` is only for internal use.

Sources/Turf/Turf.swift
RomainQuidet: Done as LineString extension in Spline.swift

Sources/Turf/Spline.swift
RomainQuidet: of course, done.

RomainQuidet:
@frederoni I updated the code according to your reviews. Thanks.

Sources/Turf/Spline.swift
```diff
@@ -0,0 +1,147 @@
...
+    }
+}
+
+public extension LineString {
+	
+	/**
+	Returns a new LineString based on bezier transformation of the input line
+	*/
+	public func bezier(resolution: Int = 10000, sharpness: Double = 0.85) -> LineString? {
+		// ported from https://github.com/Turfjs/turf/blob/master/packages/turf-bezier-spline/index.ts
```
1ec5: It’s important to use a permalink here – the URL should name either a commit or a tag. Since this file hasn’t changed since March, I assume this is what you ported:

```suggestion
 		// ported from https://github.com/Turfjs/turf/blob/1ea264853e1be7469c8b7d2795651c9114a069aa/packages/turf-bezier-spline/index.ts
```

Sources/Turf/Spline.swift
```diff
@@ -0,0 +1,147 @@
...
+		let points = coordinates.map {
+			SplinePoint(coordinate: $0)
```
1ec5: FYI, this method seems to be the only part of the PR that indents lines using tabs. (I don’t normally nitpick about whitespace, but this seemed unintentional to me.)

Sources/Turf/Spline.swift
```diff
@@ -0,0 +1,147 @@
...
+		}
+		guard let spline = Spline(points: points, duration: resolution, sharpness: sharpness) else { return nil }
+		var coords = [CLLocationCoordinate2D]()
+		for i in stride(from: 0, to: resolution, by: 10) {
+			let pos = spline.pos(time: i)
+			let index = Int(floor(Double(i) / 100))
+			if index % 2 == 0 {
+				coords.append(pos.coordinate)
```
1ec5: This is fine, but FYI there’s a more succinct, functional way to write this loop:

```swift
let coords = stride(from: 0, to: resolution, by: 10)
    .filter { Int(floor(Double($0) / 100)) % 2 == 0 }
    .map { spline.pos(time: $0) }
```

README.md
```diff
@@ -86,6 +86,7 @@ Turf.js | Turf-swift
 [turf-line-slice](https://github.com/Turfjs/turf/tree/master/packages/turf-line-slice/) | `LineString.sliced(from:to:)`
 [turf-line-slice-along](https://github.com/Turfjs/turf/tree/master/packages/turf-line-slice-along/) | `LineString.trimmed(from:distance:)`
 [turf-nearest-point-on-line](https://github.com/Turfjs/turf/tree/master/packages/turf-nearest-point-on-line/) | `LineString.closestCoordinate(to:)`
+[turf-bezier-spline](https://github.com/Turfjs/turf/tree/master/packages/turf-bezier-spline/) | `LineString.bezier(resolution:sharpness:)`
```
1ec5: Thanks for remembering to update the readme. For the reader’s convenience, can you alphabetize it below turf-area?

Sources/Turf/Spline.swift
```diff
@@ -0,0 +1,147 @@
...
+                (step.y - lastStep.y) * (step.y - lastStep.y) +
+                (step.z - lastStep.z) * (step.z - lastStep.z))
+            if dist > stepLength {
+                steps.append(t)
+                lastStep = step
+            }
+        }
+    }
+    
+    func pos(time: Int) -> SplinePoint {
```
1ec5: Nit: rename the method `position(at:)`:

```suggestion
    func position(at time: Int) -> SplinePoint {
```

Sources/Turf/Spline.swift
```diff
@@ -0,0 +1,147 @@
...
+        var t = time - delay
+        if t < 0 {
+            t = 0
```
1ec5: Or more succinctly:

```swift
var t = max(0, time - delay)
```

Sources/Turf/Spline.swift
```diff
@@ -0,0 +1,147 @@
...
+        }
+        if t > duration {
+            t = duration - 1
```
1ec5: Is it just me, or is there an off-by-one error in [the original JavaScript version](https://github.com/Turfjs/turf/blob/1ea264853e1be7469c8b7d2795651c9114a069aa/packages/turf-bezier-spline/lib/spline.ts#L131)? If `t` is equal to `duration`, then leave it alone, but if it’s greater than `duration`, cap it at `duration`−1. 🤔

Without further investigation, I’m not sure we should deviate from Turf.js, given our experience with another seemingly accidental off-by-one error that turned out to be significant: #27.

Sources/Turf/Spline.swift
```diff
@@ -0,0 +1,147 @@
...
+    init?(points: [SplinePoint], duration: Int, sharpness: Double, stepLength: Double = 60) {
+        guard points.count >= 2 else { return nil }
+        self.points = points
+        self.duration = duration
+        self.sharpness = sharpness
+        self.stepLength = stepLength
+        
+        length = points.count
+        
+        for index in stride(from: 0, to: points.count - 1, by: 1) {
```
1ec5: A stride of 1 is equivalent to a range:

```swift
center = (0..<points.count).map {
    let points = points[$0]
    …
    return center
}
```

or:

```swift
center = points.enumerated().map {
    let point = $0.1
    let nextPoint = points[$0.0 + 1]
    …
    return center
}
```

A more concise, though possibly less memory-efficient way of writing this loop would be to use `zip(_:_:)`:

```swift
zip(points, points.suffix(from: 1)).map {
    return SplinePoint(x: ($0.x + $1.x) / 2,
                       y: ($0.y + $1.y) / 2,
                       z: ($0.z + $1.z) / 2)
}
```

Sources/Turf/Spline.swift
```diff
@@ -0,0 +1,147 @@
...
+            let point = points[index]
+            let nextPoint = points[index + 1]
+            let center = SplinePoint(x: (point.x + nextPoint.x) / 2, y: (point.y + nextPoint.y) / 2, z: (point.z + nextPoint.z) / 2)
+            centers.append(center)
+        }
+        
+        controls.append((points[0], points[0]))
+        
+        for index in stride(from: 0, to: centers.count - 1, by: 1) {
```
1ec5: Similarly, this is equivalent to a range. I probably wouldn’t use `zip(_:_:)` here because we’re juggling `nextPoint` in addition to `center` and `nextCenter`.

Sources/Turf/Spline.swift
RomainQuidet: Done

Sources/Turf/Spline.swift
RomainQuidet: Done

Sources/Turf/Spline.swift
RomainQuidet: Done, I was not using swifty code to follow original code and allow easier code review / update.

README.md
RomainQuidet: My bad, done

Sources/Turf/Spline.swift
RomainQuidet: Done

Sources/Turf/Spline.swift
RomainQuidet: I updated with a range, zip could be an issue for perf on huge lines.

1ec5: Yeah, I think you made the right call. 👍

Sources/Turf/Spline.swift
1ec5: Thanks for the change, and I appreciate your effort to make the port easier to follow. 🙂

Sources/Turf/Spline.swift
1ec5: Filed upstream for future investigation: Turfjs/turf#1567.

PR #75, opened 2018-12-18
Added midpoint func

stonetip:
Modeled after midpoint function in Turfjs (https://github.com/Turfjs/turf/blob/17002ccd57e04e84ddb38d7e3ac8ede35b019c58/packages/turf-midpoint/index.js). See also, http://turfjs.org/docs#midpoint

I've tested this against several coordinate pairs and get the same results as with Turfjs (but I wouldn't say it was exhaustive). It would be helpful to have this function to mirror what's in Turfjs.


Sources/Turf/CoreLocation.swift
```diff
@@ -93,5 +93,12 @@ extension CLLocationCoordinate2D: Equatable {
     public func distance(to coordinate: CLLocationCoordinate2D) -> CLLocationDistance {
         return RadianCoordinate2D(self).distance(to: RadianCoordinate2D(coordinate)) * metersPerRadian
     }
+    
+    /// Returns the point midway between two coordinates measured in degrees
+    public func midpoint(to coordinate: CLLocationCoordinate2D) -> CLLocationCoordinate2D{
```
frederoni: this should be moved to `Point.midpoint(to:)` in `Point.swift`.

frederoni:
Thanks for porting this feature. Would you mind porting the tests as well?

Sources/Turf/CoreLocation.swift
stonetip: I'll move it there.

stonetip:
I'll add tests. Assuming I should port them from https://github.com/Turfjs/turf/blob/master/packages/turf-midpoint/test.js

stonetip:
Note: I changed the func signature to take both end points to mirror the Turfjs midpoint func, which has as its input params:

 * @param {Coord} point1 first point
 * @param {Coord} point2 second point

Sources/Turf/Point.swift
```diff
@@ -41,3 +41,11 @@ public struct PointFeature: GeoJSONObject {
         try container.encodeIfPresent(identifier, forKey: .identifier)
     }
 }
+
+
+/// Returns the point midway between two coordinates measured in degrees
+public func midpoint(point1: CLLocationCoordinate2D, point2: CLLocationCoordinate2D) -> CLLocationCoordinate2D{
```
1ec5: Can you move this freestanding function into Turf.swift? So far, we’ve defined static methods in the Turf struct instead of freestanding functions, to avoid polluting the main namespace. That’s somewhat redundant to using the module name as a namespace, but it’s good to be consistent.

> I changed the func signature to take both end points to mirror the Turfjs midpoint func

As [the readme](https://github.com/mapbox/turf-swift#available-functionality) illustrates, this library doesn’t necessarily mirror Turf.js function signatures, because doing so results in very non-idiomatic Swift code. Following the [Swift API design guidelines](https://swift.org/documentation/api-design-guidelines/) is more important.

I’d normally be inclined to ask for argument labels, but [this guideline](https://swift.org/documentation/api-design-guidelines/#no-labels-for-indistinguishable-arguments) makes an exception for cases where the two arguments are interchangeable, as they are in `midpoint(_:_:)`.

By the way, this method is called “midpoint” but returns a coordinate instead of a point geometry or feature. A coordinate is more convenient; on the other hand, the JavaScript version returns a point feature. I think it’s OK, but we may end up needing to rename this method in the future if we decide to make one that returns a point feature.

stonetip: To maintain some consistency with Turfjs, keep the name intuitive, and avoid conflict with future code that might return a point instead of CLLocationCoordinate2D, I renamed the func to mid.

Within the func I made the argument labels omittable.

I hope that I placed the func in the right place, in the right way. Any guidance (such as the ref to the API design guidelines) is much appreciated. My goal is to contribute a func a week.

1ec5:
Added to the readme in 86a1b6bcc9d8b88dfaf6066ed6d0cdce761003cf.

PR #90, opened 2020-01-09
Fix Polygon.contains

### LINKED ISSUE #88: Polygon.contains(_:ignoreBoundary:) returns false for coordinates on boundary coinciding with bounding box
When specifying `ignoreBoundary: false` in `Polygon.contains(_:ignoreBoundary:)` (the default value), coordinates on the boundary that coincide with the bounding box are incorrectly considered not to be contained.

`Ring.contains(_:ignoreBoundary:)` on the other hand works as expected.

The issue is that the optimization used in `Polygon.contains(_:ignoreBoundary:)` to first do a check whether the point is within the bounding box

https://github.com/mapbox/turf-swift/blob/22246d45d321006b938a2102283efa6911ce1e66/Sources/Turf/Polygon.swift#L73-L76

does not pass on `ignoreBoundary` as `BoundingBox.contains(_:)` doesn't have this parameter:

https://github.com/mapbox/turf-swift/blob/22246d45d321006b938a2102283efa6911ce1e66/Sources/Turf/BoundingBox.swift#L30

The following code shows the problem:

```swift
let center = CLLocationCoordinate2D(latitude:  0, longitude:  0)
let sw     = CLLocationCoordinate2D(latitude: -1, longitude: -1)
let se     = CLLocationCoordinate2D(latitude: -1, longitude:  1)
let ne     = CLLocationCoordinate2D(latitude:  1, longitude:  1)
let nw     = CLLocationCoordinate2D(latitude:  1, longitude: -1)

let coords = [sw, nw, ne, se, sw]

let polygon = Polygon([coords])
let ring = Ring(coordinates: coords)

// Center is contained:

XCTAssertTrue(ring.contains(center))
XCTAssertTrue(polygon.contains(center))

// Corners are not contained when excluding boundary:

XCTAssertFalse(ring.contains(sw, ignoreBoundary: true))
XCTAssertFalse(ring.contains(nw, ignoreBoundary: true))
XCTAssertFalse(ring.contains(ne, ignoreBoundary: true))
XCTAssertFalse(ring.contains(se, ignoreBoundary: true))

XCTAssertFalse(polygon.contains(sw, ignoreBoundary: true))
XCTAssertFalse(polygon.contains(nw, ignoreBoundary: true))
XCTAssertFalse(polygon.contains(ne, ignoreBoundary: true))
XCTAssertFalse(polygon.contains(se, ignoreBoundary: true))

// Corners are contained when including boundary:

XCTAssertTrue(ring.contains(sw, ignoreBoundary: false))
XCTAssertTrue(ring.contains(nw, ignoreBoundary: false))
XCTAssertTrue(ring.contains(ne, ignoreBoundary: false))
XCTAssertTrue(ring.contains(se, ignoreBoundary: false))

XCTAssertTrue(polygon.contains(sw, ignoreBoundary: false)) // XCTAssertTrue failed
XCTAssertTrue(polygon.contains(nw, ignoreBoundary: false)) // XCTAssertTrue failed
XCTAssertTrue(polygon.contains(ne, ignoreBoundary: false)) // XCTAssertTrue failed
XCTAssertTrue(polygon.contains(se, ignoreBoundary: false)) // XCTAssertTrue failed
```

I'd suggest to address this in the following way:

- Change `BoundingBox.contains(_:)` to `BoundingBox.contains(_:ignoreBoundary:)` with a default of `true` to preserve the current behavior.
- Optionally, move the bounding box optimization from `Polygon` to `Ring` in order for `Ring` to benefit from this optimization as well (`Polygon` would still benefit indirectly from this optimization).

I'd be happy to make a PR if you agree on the suggested approach.

fphilipe:
Fixes #88

-   **Fix `Polygon.contains` at boundary**

    `Polygon.contains(_:ignoreBoundary:)` with `ignoreBoundary` set to `false` (the default), was returning `false` for a point at the boundary that coincides with the bounding box. This is because the optimization that first checks whether the point is contained by the polygon's bounding box, which uses `BoundingBox.contains(_:)`, did not consider a point at the boundary as being contained.

    To address this, `BoundingBox.contains(_:)` is extended with a second argument `ignoreBoundary` defaulting to `true` for backwards compatibility.  The optimization in `Polygon.contains(_:ignoreBoundary:)` now forwards the `ignoreBoundary` argument, fixing the issue.

-   **Move containment optimization from `Polygon` to `Ring`**

    `Polygon.contains(_:ignoreBoundary:)` first checks whether its bounding box contains the point before proceeding with a more expensive and complete check. For polygons with holes, each hole could benefit from this optimization. In other words, the outer ring as well as all inner rings can benefit from this optimization.

    This commit accomplishes this by moving the bounding box optimization to `Ring.contains(_:ignoreBoundary:)`.

-   **Fix `Polygon.contains` at boundary of hole**

    When including the boundary, a point at the boundary of a hole is considered to be contained by the polygon. Conversely, when not including the boundary a point at the boundary of a hole is considered not to be contained by the polygon.

    Neither of those were the case as shown by the added test. The fix is as simple as flipping the `ignoreBoundary` flag when checking whether the point is within one of the holes.

### DIFF
--- a/Sources/Turf/BoundingBox.swift
+++ b/Sources/Turf/BoundingBox.swift
@@ -27,11 +27,18 @@ public struct BoundingBox: Codable {
         self.southEast = southEast
     }
     
-    public func contains(_ coordinate: CLLocationCoordinate2D) -> Bool {
-        return southEast.latitude < coordinate.latitude
-            && northWest.latitude > coordinate.latitude
-            && northWest.longitude < coordinate.longitude
-            && southEast.longitude > coordinate.longitude
+    public func contains(_ coordinate: CLLocationCoordinate2D, ignoreBoundary: Bool = true) -> Bool {
+        if ignoreBoundary {
+            return southEast.latitude < coordinate.latitude
+                && northWest.latitude > coordinate.latitude
+                && northWest.longitude < coordinate.longitude
+                && southEast.longitude > coordinate.longitude
+        } else {
+            return southEast.latitude <= coordinate.latitude
+                && northWest.latitude >= coordinate.latitude
+                && northWest.longitude <= coordinate.longitude
+                && southEast.longitude >= coordinate.longitude
+        }
     }
     
     // MARK: - Codable
--- a/Sources/Turf/Polygon.swift
+++ b/Sources/Turf/Polygon.swift
@@ -32,10 +32,6 @@ extension Geometry.PolygonRepresentation {
     ///
     ///Ported from: https://github.com/Turfjs/turf/blob/e53677b0931da9e38bb947da448ee7404adc369d/packages/turf-boolean-point-in-polygon/index.ts#L31-L75
     public func contains(_ coordinate: CLLocationCoordinate2D, ignoreBoundary: Bool = false) -> Bool {
-        let bbox = BoundingBox(from: coordinates.first)
-        guard bbox?.contains(coordinate, ignoreBoundary: ignoreBoundary) ?? false else {
-            return false
-        }
         guard outerRing.contains(coordinate, ignoreBoundary: ignoreBoundary) else {
             return false
         }
--- a/Sources/Turf/Polygon.swift
+++ b/Sources/Turf/Polygon.swift
@@ -36,7 +36,7 @@ extension Geometry.PolygonRepresentation {
             return false
         }
         for ring in innerRings {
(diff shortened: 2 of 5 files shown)


fphilipe:
Friendly ping 😊 Let me know if this is something that is of interest to you. Otherwise we can close it also.

Issue #91, opened 2020-01-12
Associated-value enumeration should be for geometries, not features


1ec5:
The GeoJSON model types in this library currently mirror the shape classes in the map SDK, which are inconsistent with the GeoJSON specification in several significant ways. In particular, there are multiple feature structs, one for each geometry type, and no single type unifies the various geometry types. Aside from nonconformance with GeoJSON, this approach makes it difficult to support GeometryCollection (#89).

* [x] Replace the FeatureVariant associated-value enumeration with a single struct Feature that contains a Geometry
* [x] Replace the Geometry struct with an associated-value enumeration
* [x] Remove the various *Feature structs in favor of Feature

/ref mapbox/mapbox-gl-native#7454
/cc @mapbox/navigation-ios @frederoni

1ec5:
Fixed by #93.

PR #93, opened 2020-02-26
GeoJSON Geometry Collections support

### LINKED ISSUE #89: Missing GeometryCollection support
Will this be supported eventually, or was it intentionally left out?

Udumft:
WORK-IN-PROGRESS

Resolves #89 and #91


Sources/Turf/FeatureCollection.swift
```diff
@@ -17,21 +18,23 @@ public struct FeatureCollection: GeoJSONObject {
...
-                .lineStringFeature(LineStringFeature(LineString($0[0])))
-        }
-    }
+//    public init(_ multiPolygon: MultiPolygon) {
+//        self.features = multiPolygon.coordinates.map {
+//            $0.count > 1 ?
+//                .multiLineStringFeature(.init(.init($0))) :
+//                .lineStringFeature(LineStringFeature(LineString($0[0])))
+//        }
+//    }
```
Udumft: Clarifications about this initializer would be appreciated)

Sources/Turf/GeoJSON.swift
```diff
@@ -121,10 +120,12 @@ public struct Feature: Codable {
     }
 }
 
-public enum GeoJSONType: String {
+public enum GeoJSONType: String, CaseIterable {
```
Udumft: I suggest to use `CaseIterable` instead of manually defining `allValues`

Sources/Turf/GeoJSON.swift
```diff
@@ -200,7 +185,7 @@ public class GeoJSON: Codable {
      
      - Returns: decoded GeoJSON of type T.
      */
-    public static func parse<T: GeoJSONObject>(_ type: T.Type, from data: Data) throws -> T {
+    public static func parse<T: GeoJSONObject>(_ type: T.Type, from data: Data) throws -> T {  // ???
```
Udumft: Replacing separate structures for each Feature with a single base one will make this method almost useless...

Sources/Turf/Geometry.swift
```diff
@@ -4,21 +4,129 @@ import CoreLocation
...
+    
+    case Point(coordinates: CLLocationCoordinate2D)
+    case LineString(coordinates: [CLLocationCoordinate2D])
+    case Polygon(coordinates: [[CLLocationCoordinate2D]])
+    case MultiPoint(coordinates: [CLLocationCoordinate2D])
+    case MultiLineString(coordinates: [[CLLocationCoordinate2D]])
+    case MultiPolygon(coordinates: [[[CLLocationCoordinate2D]]])
+    case GeometryCollection(geometries: [_Geometry])
+    
+    public var type: GeometryType {
```
Udumft: Having all these cases combined in a single `Enum` will lead to a lot of switch cases in the code... does not seem pretty to me. Am I missing something?

Sources/Turf/Geometry.swift
```diff
@@ -4,21 +4,129 @@ import CoreLocation
...
+            return value
+        case .MultiPolygon(let value):
+            return value
+        case .GeometryCollection(let value):
+            return value
+        }
     }
 }
+
+extension _Geometry: Codable {
```
Udumft: Also, I see that some "Geometry" structs had additional methods (like searching for intersections, additional initializers and so on). To maintain these functionality, such methods will have to be moved inside this Enum or modified to become a free functions. How can I make this better than this?

Tests/TurfTests/GeoJSONTests.swift
```diff
@@ -6,9 +6,11 @@ class GeoJSONTests: XCTestCase {
...
     func testPoint() {
         let coordinate = CLLocationCoordinate2D(latitude: 10, longitude: 30)
-        let point = Point(coordinate)
-        let pointFeature = PointFeature(point)
-        XCTAssertEqual(pointFeature.geometry.coordinates, coordinate)
+        let geometry = _Geometry.Point(coordinates: coordinate)
+        let pointFeature = _Feature(geometry)
+        
+//        XCTAssertEqual(pointFeature.geometry.coordinates, coordinate)
+        XCTAssertEqual(pointFeature.geometry.value as! CLLocationCoordinate2D, coordinate)
```
Udumft: Using enum with associated value makes `Feature` struct usage a bit clumsy, especially in comparison with previous implementation.

Udumft:
I understand that this is not meant to be backward-compatible, but still tried to make it as close as possible.But at some points it looks like it's a complete API replacement)

Sources/Turf/Geometry.swift
```diff
@@ -4,21 +4,129 @@ import CoreLocation
...
-public struct Geometry: Codable {
-    public var type: String
+public enum _Geometry {
+    private enum CodingKeys: String, CodingKey {
+        case type
+        case coordinates
+        case geometries
+    }
+    
+    case Point(coordinates: CLLocationCoordinate2D)
```
JThramer: `coordinate`, not `coordinates`

1ec5:
> I understand that this is not meant to be backward-compatible, but still tried to make it as close as possible.But at some points it looks like it's a complete API replacement)

It’s OK to replace the API if it would be more expedient to do so, with less hair-pulling. 😄

Sources/Turf/GeoJSON.swift
frederoni: Yes, `CaseIterable` is much cleaner but it wasn't available until Swift 5.0, which explains why the Linux CI bot on Swift 4.1 fails. Feel free to bump the requirement to 5+ in the SPM manifest.

Sources/Turf/Geometry.swift
1ec5: One possibility would be to have each case in the Geometry enumeration contain a struct that represents the actual geometry and can perform operations specific to that kind of geometry. For example, `case .Polygon(polygon: Polygon)`. Not sure if that would make things easier though.

In Turf.js, everything is a free function, but we originally wanted a more object-oriented API in Swift. Turf.js functions generally know how to return a meaningful result for any kind of geometry, so I think it could be OK to fold these methods into Geometry as long as we similarly generalize them.

Sources/Turf/Geometry.swift
Udumft: Also, original `struct Point` refers to it's property as `coordinates`. So I vote for "coordinate**s**" :)

Sources/Turf/Geometry.swift
```diff
@@ -4,21 +4,157 @@ import CoreLocation
...
+    public var multiPolygon: [[[CLLocationCoordinate2D]]]? {
+        guard case let .MultiPolygon(coordinates: coordinates) = self else { return nil }
+        return coordinates
+    }
+    /// Returns geometries collection if current enum case is `.GeometryCollection`
+    public var geometryCollection: [Geometry]? {
+        guard case let .GeometryCollection(geometries: geometries) = self else { return nil }
+        return geometries
+    }
+}
```
JThramer: Please delete these. The idea behind an associated-value enum is to force the developer to deal with variants at point of use instead of using "magic optionals" that may or may not be nil depending on the variance.

If a developer expects a _certain_ case in an enum, they can quickly resolve for it by using `guard case let .GeometryCondition(coordinates) = enum else { panic() }`
```suggestion

```

README.md
```diff
@@ -107,7 +107,7 @@ let geojson = try! GeoJSON.parse(data: data, as: FeatureCollection.self)
 
 // Initialize a PointFeature and encode as GeoJSON
 let coordinate = CLLocationCoordinate2D(latitude: 0, longitude: 1)
-let point = Geometry.Point(coordinates: coordinate)
+let point = Geometry.Point(coordinates: .init(coordinate))
```
JThramer: Why was this added? You're creating a `CLLocationCoordinate2D` from a `CLLocationCoordinate2D` it seems.

Sources/Turf/Feature.swift
```diff
@@ -0,0 +1,38 @@
+import Foundation
+#if !os(Linux)
+import CoreLocation
+#endif
+
+
+public struct Feature: GeoJSONObject {
+    public var type: FeatureType = .feature
+    public var identifier: FeatureIdentifier?
+    public var properties: [String : AnyJSONType]?
```
JThramer: I know this pattern was introduced before this PR, but I feel like `[String: AnyJSONType]` is a code-smell that we should remove ASAP. The `AnyJSONType` class doesn't really buy you anything -- You'd still have to cast `someAnyJSONTypeObject.jsonValue as? String` to get the correct type-safety.

If that's the case, we may as well represent JSON arrays as `[String: Any]`, and let developers conditionally-cast for the type they're looking for, as needed. `let someDouble = someJSONValue as? Double` is a very familiar pattern in swift API's.

However, this may be out-of-scope for this PR, and may be better served in a fast follow. @1ec5 @frederoni what say you?

1ec5: I think `AnyJSONType` was first introduced to facilitate decoding, but we only support JSONDecoder anyways, so we can be sure that decoding a `[String: Any?]` dictionary would contain only JSON-representable values. (Note that it should be `[String: Any?]`, not `[String: Any]`, because `null` is a valid JSON type.)

@JThramer, can you open a separate issue tracking that issue? I think @Udumft can make progress without having to consider this issue for now.

JThramer: Groovy, I'll go ahead and approve and cut the FF ticket.

JThramer: Ticket opened: https://github.com/mapbox/turf-swift/issues/94

PR #100, opened 2020-06-02
Apply Swift naming conventions to GeoJSON initializers

1ec5:
This PR makes a couple changes to make the GeoJSON interface read better in Swift:

* Moved argument labels to GeoJSON struct initializers. It was misleading to label a geometry-typed associated value as `coordinates`. Labeling the `Geometry`-typed argument to `Feature`’s initializer means less confusion as we continue to add convenience initializers.
* Lowercased the first letter of each of `Geometry`’s cases, per Swift naming conventions. (GeoJSON always capitalizes these terms, but Swift lowercases capitalized terms at the beginning of case names anyways, like `.url` and `.mapbox`.)

For example:

```swift
// Before
Feature(.Point(coordinates: Point(coordinate)))
// After
Feature(geometry: .point(Point(coordinate)))
```

Some errors have been corrected in the example code in the readme.

/cc @mapbox/navigation-ios @frederoni

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ let geojson = try! GeoJSON.parse(FeatureCollection.self, from: data)
 // Initialize a PointFeature and encode as GeoJSON
 let coordinate = CLLocationCoordinate2D(latitude: 0, longitude: 1)
 let point = Point(coordinate)
-let pointFeature = Feature(geometry: .Point(point))
+let pointFeature = Feature(geometry: .point(point))
 let data = try! JSONEncoder().encode(pointFeature)
 let json = String(data: data, encoding: .utf8)
 print(json)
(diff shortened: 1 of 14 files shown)


1ec5:
Travis and Bitrise builds are failing:

```
/home/travis/build/mapbox/turf-swift/Sources/Turf/Geometries/GeometryCollection.swift:17:18: error: enum type 'Geometry' has no case 'MultiLineString'; did you mean 'multiLineString'

                .MultiLineString(coordinates: .init($0)) :

                 ^~~~~~~~~~~~~~~

                 multiLineString
```

This code is coming from master rather than this PR’s branch. I think what’s going on is that TurfTests depends on Turf, and Package.swift specifies Turf as the contents of master.

PR #106, opened 2020-06-10
[Multi]Polygon convenience additions

Udumft:
Based on [community request](https://github.com/mapbox/turf-swift/pull/92) with updating to new type system.
Added convenience initializers and methods for Polygon and MultiPolygon. 
Tests included.

### DIFF
--- a/Sources/Turf/Geometries/MultiPolygon.swift
+++ b/Sources/Turf/Geometries/MultiPolygon.swift
@@ -10,4 +10,32 @@ public struct MultiPolygon: Equatable {
     public init(_ coordinates: [[[CLLocationCoordinate2D]]]) {
         self.coordinates = coordinates
     }
+    
+    public init(_ polygons: [Polygon]) {
+        self.coordinates = polygons.map { (polygon) -> [[CLLocationCoordinate2D]] in
+            return polygon.coordinates
+        }
+    }
+}
+
+extension MultiPolygon {
+    
+    public var polygons: [Polygon] {
+        return coordinates.map { (coordinates) -> Polygon in
+            return Polygon(coordinates)
+        }
+    }
+    
+    /**
+     * Determines if the given coordinate falls within any of the polygons.
+     * The optional parameter `ignoreBoundary` will result in the method returning true if the given coordinate
+     * lies on the boundary line of the polygon or its interior rings.
+     *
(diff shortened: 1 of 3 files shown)


Sources/Turf/Geometries/MultiPolygon.swift
```diff
@@ -10,4 +10,32 @@ public struct MultiPolygon: Equatable {
...
+            return Polygon(coordinates)
+        }
+    }
+    
+    /**
+     * Determines if the given coordinate falls within any of the polygons.
+     * The optional parameter `ignoreBoundary` will result in the method returning true if the given coordinate
+     * lies on the boundary line of the polygon or its interior rings.
+     *
+     * Calls contains funcion for each contained polygon
```
1ec5, line 34: Typo:

```suggestion
     * Calls contains function for each contained polygon
```

Sources/Turf/Geometries/MultiPolygon.swift
```diff
@@ -10,4 +10,32 @@ public struct MultiPolygon: Equatable {
...
+     */
+    public func contains(_ coordinate: CLLocationCoordinate2D, ignoreBoundary: Bool = false) -> Bool {
+        return polygons.first(where: {
+            $0.contains(coordinate, ignoreBoundary: ignoreBoundary)
+        }) != nil
```
1ec5, line 39: Minor performance optimization:

```suggestion
        return polygons.contains {
            $0.contains(coordinate, ignoreBoundary: ignoreBoundary)
        }
```

Sources/Turf/Geometries/Polygon.swift
```diff
@@ -10,6 +10,15 @@ public struct Polygon: Equatable {
...
         self.coordinates = coordinates
     }
+    
+    public init(outerRing: Ring, innerRings: [Ring] = []) {
+        var ringCoordinates = [outerRing.coordinates]
+        for innerRing in innerRings {
+            ringCoordinates.append(innerRing.coordinates)
+        }
+        
+        self.coordinates = ringCoordinates
```
1ec5, line 20: Nit: I didn’t try this, but it should be possible to map directly to the coordinates thanks to the default argument value above:

```suggestion
        coordinates = ([outerRing] + innerRings).map { $0.coordinates }
```

PR #113, opened 2020-07-21
Update README.md

captainbarbosa:
Updates some links in the readme as a result of changing the primary branch name from `master` to `main`.


captainbarbosa:
It looks like the branch change didn't affect CI, which is also good 👍

PR #114, opened 2020-08-25
Document plurality of Point geometry

captainbarbosa:
~Changes the initializer for `Point` so that the argument label and parameter is `coordinate` instead of `coordinates` for grammatical consistency.~ This is part of the GeoJSON specification, so I've changed this PR to just clarify that in the docs.

### DIFF
--- a/Sources/Turf/Geometries/Point.swift
+++ b/Sources/Turf/Geometries/Point.swift
@@ -5,6 +5,9 @@ import CoreLocation
 
 
 public struct Point: Equatable {
+    /** Note: The pluralization of `coordinates` is defined
+     in the GeoJSON RFC, so we've kept it for consistency.
+     https://tools.ietf.org/html/rfc7946#section-1.5 */
     public var coordinates: CLLocationCoordinate2D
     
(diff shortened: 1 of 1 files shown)


Sources/Turf/Geometries/Point.swift
```diff
@@ -5,9 +5,9 @@ import CoreLocation
 
 
 public struct Point: Equatable {
-    public var coordinates: CLLocationCoordinate2D
+    public var coordinate: CLLocationCoordinate2D
```
frederoni, line 8: The reasoning behind this name in the first place was that according to the [GeoJSON RFC](https://tools.ietf.org/html/rfc7946#section-1.5), a Point’s coordinate is serialized as `coordinates` though everything but the first coordinate is being ignored.

captainbarbosa: Oh, okay, that makes sense. I've reset this PR and added that information to the incline documentation so others might not try to do it again in the future. How does that sound?

frederoni (approved):

PR #122, opened 2020-11-13
Update README.md to mention Turf.js bearing

captainbarbosa:
Adds link to turf bearing docs to readme.


1ec5 (approved):
Thanks!

1ec5:
I guess we can realphabetize this table according to the first column, but it’s no big deal.

PR #123, opened 2020-11-15
Add circle function

captainbarbosa:
Adds Turf's [circle function](https://turfjs.org/docs/#circle) to return a polygon that approximates a circle with a given radius.

**Usage:**

```swift
let centerCoordinate = CLLocationCoordinate2D(latitude: 20, longitude: -87)
let radius = CLLocationDistance(radius)
let circleShape = Polygon(from: centerCoordinate, radius: radius)
```

Below is a visual representation built with the Mapbox Maps SDK, which the vertices are marked as dots:

</img>

### DIFF
--- a/Sources/Turf/Geometries/Polygon.swift
+++ b/Sources/Turf/Geometries/Polygon.swift
@@ -14,6 +14,28 @@ public struct Polygon: Equatable {
     public init(outerRing: Ring, innerRings: [Ring] = []) {
         self.coordinates = ([outerRing] + innerRings).map { $0.coordinates }
     }
+
+    /**
+     Returns a circular polygon based on a given radius and a defined
+     number of vertices for precision.
+
+     - Parameter center: The center coordinate for the circle.
+     - Parameter radius: The radius of the circle, measured in meters.
+     - Parameter vertices: The number of vertices the circle will have. Defaults to 64.
+     - Returns: A polygon resembling a circle.
+     */
+    public init(center: CLLocationCoordinate2D, radius: CLLocationDistance, vertices: Int = 64) {
+        // The first and last coordinates in a polygon must be identical,
+        // which is why we're using the inclusive range operator in this case.
+        // Ported from https://github.com/Turfjs/turf/blob/17002ccd57e04e84ddb38d7e3ac8ede35b019c58/packages/turf-circle/index.ts
+        let coordinates = (0...vertices).map { ( step ) -> CLLocationCoordinate2D in
+            let bearing = fabs(CLLocationDirection(step * -360 / vertices))
+            print("bearing: \(bearing)")
+            return center.coordinate(at: radius, facing: bearing)
+        }
+
+        self.coordinates = [coordinates]
+    }
 }
 
 extension Polygon {
--- a/Sources/Turf/Geometries/Polygon.swift
+++ b/Sources/Turf/Geometries/Polygon.swift
@@ -21,10 +21,11 @@ public struct Polygon: Equatable {
 
      - Parameter center: The center coordinate for the polygon.
      - Parameter radius: The radius of the polygon, measured in meters.
-     - Parameter vertices: The number of vertices the polygon will have. Defaults to 64.
+     - Parameter vertices: The number of vertices the polygon will have.
+                           The recommended amount is 64.
      - Returns: A polygon shape which approximates a circle.
      */
-    public init(center: CLLocationCoordinate2D, radius: CLLocationDistance, vertices: Int = 64) {
+    public init(center: CLLocationCoordinate2D, radius: CLLocationDistance, vertices: Int) {
         // The first and last coordinates in a polygon must be identical,
         // which is why we're using the inclusive range operator in this case.
         // Ported from https://github.com/Turfjs/turf/blob/17002ccd57e04e84ddb38d7e3ac8ede35b019c58/packages/turf-circle/index.ts
--- a/Sources/Turf/Geometries/Polygon.swift
+++ b/Sources/Turf/Geometries/Polygon.swift
@@ -16,13 +16,13 @@ public struct Polygon: Equatable {
     }
 
     /**
-     Returns a circular polygon based on a given radius and a defined
-     number of vertices for precision.
+     Initializes a polygon as a given center coordinate with a given number of
+     vertices, as a means to approximate a circle.
 
-     - Parameter center: The center coordinate for the circle.
-     - Parameter radius: The radius of the circle, measured in meters.
-     - Parameter vertices: The number of vertices the circle will have. Defaults to 64.
-     - Returns: A polygon resembling a circle.
+     - Parameter center: The center coordinate for the polygon.
+     - Parameter radius: The radius of the polygon, measured in meters.
+     - Parameter vertices: The number of vertices the polygon will have. Defaults to 64.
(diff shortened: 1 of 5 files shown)


Sources/Turf/Turf.swift
```diff
@@ -46,3 +46,21 @@ public func mid(_ coord1: CLLocationCoordinate2D, _ coord2: CLLocationCoordinate
...
+
+/**
+ Takes a coordinate and calculates the circular polygon given a radius in meters,
+ using a defined number of steps for precision.
+
+ - Parameter center: The center coordinate for the circle.
+ - Parameter radius: The radius of the circle, measured in meters.
+ - Parameter steps: The number of vertices the circle will have. Defaults to 64.
+ */
+public func circle(from center: CLLocationCoordinate2D, radius: CLLocationDistance, steps: Int = 64) -> Polygon {
```
1ec5, line 58: Consider turning this freestanding function into a Polygon initializer: `Polygon(center:radius:vertices:)`.

Sources/Turf/Turf.swift
```diff
@@ -46,3 +46,21 @@ public func mid(_ coord1: CLLocationCoordinate2D, _ coord2: CLLocationCoordinate
...
+
+    let coordinates = (0...steps).map { ( step ) -> CLLocationCoordinate2D in
```
1ec5, line 60: 👍 This is correct, because the GeoJSON specification requires the first and last coordinates in a linear ring to be identical. It might be worth a comment so someone doesn’t come along thinking it’s an off-by-one error and introduces an off-by-one error.

Sources/Turf/Turf.swift
```diff
@@ -46,3 +46,21 @@ public func mid(_ coord1: CLLocationCoordinate2D, _ coord2: CLLocationCoordinate
...
+/**
+ Takes a coordinate and calculates the circular polygon given a radius in meters,
+ using a defined number of steps for precision.
+
+ - Parameter center: The center coordinate for the circle.
+ - Parameter radius: The radius of the circle, measured in meters.
+ - Parameter steps: The number of vertices the circle will have. Defaults to 64.
+ */
+public func circle(from center: CLLocationCoordinate2D, radius: CLLocationDistance, steps: Int = 64) -> Polygon {
+
```
1ec5, line 59: By convention, a nontrivial port from Turf.js should have a comment containing a permalink to the source code that was ported, so we can follow up after the JavaScript implementation changes.

Sources/Turf/Turf.swift
```diff
@@ -46,3 +46,21 @@ public func mid(_ coord1: CLLocationCoordinate2D, _ coord2: CLLocationCoordinate
...
+    let coordinates = (0...steps).map { ( step ) -> CLLocationCoordinate2D in
+        let bearing = CLLocationDirection(step * -360 / steps)
```
1ec5, line 61: Is there a particular reason for going counterclockwise? Technically, CLLocationDirection uses a negative value to indicate an invalid direction. It looks like we’re getting away with it here, but there’s no guarantee that a future iteration of `coordinate(at:facing:)` won’t require the direction to be positive.

1ec5 (commented):
Thanks, this will be pretty handy!

Sources/Turf/Turf.swift
```diff
@@ -46,3 +46,21 @@ public func mid(_ coord1: CLLocationCoordinate2D, _ coord2: CLLocationCoordinate
...
+        return center.coordinate(at: radius, facing: bearing)
+    }
+
+    return Polygon([coordinates])
```
1ec5, line 65: Possible tail work: the Objective-C implementation in mapbox/mapbox-gl-native#14534 added a special case for a circle that extends to ±90° latitude, such as a hemisphere. It would be great to bring that functionality to Turf, to remove a big caveat to using this function:

https://github.com/mapbox/mapbox-gl-native/blob/910639168aaad8f06749c5be184ae50dbe9d566b/platform/darwin/src/MGLCircle.mm#L110-L118

Sources/Turf/Turf.swift
1ec5: If you do end up following the Objective-C implementation in https://github.com/mapbox/turf-swift/pull/123#discussion_r532734132, it may be necessary to go counterclockwise to patch up the polygon in the right order. Either way, it would be safer to mod the angle to [0, 360) to avoid a negative angle.

captainbarbosa: >Is there a particular reason for going counterclockwise?

I was closely following the [JS implementation](https://github.com/Turfjs/turf/blob/17002ccd57e04e84ddb38d7e3ac8ede35b019c58/packages/turf-circle/index.ts#L36) :) 

Would `fabs()` around the bearing be sufficient? That keeps everything in a positive value. I've included it in the latest change.

Sources/Turf/Geometries/Polygon.swift
```diff
@@ -14,6 +14,28 @@ public struct Polygon: Equatable {
...
+     - Parameter vertices: The number of vertices the circle will have. Defaults to 64.
+     - Returns: A polygon resembling a circle.
+     */
+    public init(center: CLLocationCoordinate2D, radius: CLLocationDistance, vertices: Int = 64) {
+        // The first and last coordinates in a polygon must be identical,
+        // which is why we're using the inclusive range operator in this case.
+        // Ported from https://github.com/Turfjs/turf/blob/17002ccd57e04e84ddb38d7e3ac8ede35b019c58/packages/turf-circle/index.ts
+        let coordinates = (0...vertices).map { ( step ) -> CLLocationCoordinate2D in
+            let bearing = fabs(CLLocationDirection(step * -360 / vertices))
+            print("bearing: \(bearing)")
```
1ec5, line 33: Remove this print statement before merging.

Sources/Turf/Turf.swift
1ec5: Yes, that will effectively make it go clockwise.

Sources/Turf/Geometries/Polygon.swift
```diff
@@ -14,6 +14,28 @@ public struct Polygon: Equatable {
...
     }
+
+    /**
+     Returns a circular polygon based on a given radius and a defined
+     number of vertices for precision.
+
+     - Parameter center: The center coordinate for the circle.
+     - Parameter radius: The radius of the circle, measured in meters.
+     - Parameter vertices: The number of vertices the circle will have. Defaults to 64.
+     - Returns: A polygon resembling a circle.
```
1ec5, line 25: It would be more accurate to describe this initializer as producing a regular polygon centered at the given coordinate. We can document it as a way to approximate a circle as well.

Sources/Turf/Geometries/Polygon.swift
```diff
@@ -14,6 +14,28 @@ public struct Polygon: Equatable {
...
+     */
+    public init(center: CLLocationCoordinate2D, radius: CLLocationDistance, vertices: Int = 64) {
```
1ec5, line 27: By describing it as a regular polygon initializer, we set expectations so that it isn’t strictly necessary to have a default value for the number of vertices. Most likely the client should explicitly choose the number of vertices anyways.

1ec5 (approved):
Just some minor feedback about how we’re positioning this new functionality, but otherwise it’s good to go on a technical level.

1ec5 (approved):

PR #124, opened 2020-12-02
Add LineString simplification

avi-c:
Port algorithm in turf-simplify to add simplification of a LineString via Ramer–Douglas–Peucker algorithm.


Sources/Turf/Geometries/LineString.swift
```diff
@@ -215,4 +215,114 @@ extension LineString {
         
         return closestCoordinate
     }
+
+    private func getSquareDistance(coordinate1: CLLocationCoordinate2D, coordinate2: CLLocationCoordinate2D) -> Double {
```
1ec5, line 219: Nit: Only include “get” in the name of an asynchronous method or one with significant side effects. Label the two coordinates `from` and `to`.

Sources/Turf/Geometries/LineString.swift
```diff
@@ -215,4 +215,114 @@ extension LineString {
...
+        let dx = coordinate1.longitude - coordinate2.longitude
+        let dy = coordinate1.latitude - coordinate2.latitude
+        return dx * dx + dy * dy
+  }
+
+    private func simplifyRadialDistance(_ coordinates: [CLLocationCoordinate2D], tolerance: Double) -> [CLLocationCoordinate2D] {
```
1ec5, line 225: The first argument isn’t a distance, so it needs an argument label. How would we describe this method in a sentence?

Sources/Turf/Geometries/LineString.swift
```diff
@@ -215,4 +215,114 @@ extension LineString {
...
+                simplifyDouglasPeuckerStep(coordinates, first: first, last: index, tolerance: tolerance, simplified: &simplified)
+            }
+            simplified.append(coordinates[index])
+            if last - index > 1 {
+                simplifyDouglasPeuckerStep(coordinates, first: index, last: last, tolerance: tolerance, simplified: &simplified)
+            }
+        }
+    }
+
+    private func simplifyDouglasPeucker(_ coordinates: [CLLocationCoordinate2D], tolerance: Double!) -> [CLLocationCoordinate2D] {
```
1ec5, line 297: Does `tolerance` need to be an implicitly unwrapped optional? We can avoid the risk of a crash by making it a `Double`, since `simplifyDouglasPeuckerStep(_:first:last:tolerance:simplified:)` takes a `Double`.

Sources/Turf/Geometries/LineString.swift
```diff
@@ -215,4 +215,114 @@ extension LineString {
...
+    public func simplified(tolerance: Double = 1.0, highestQuality: Bool = false) -> LineString {
+        guard coordinates.count > 2 else { return LineString(coordinates) }
+
+        let squareTolerance = tolerance * tolerance
+
+        var simplifiedCoordinates = highestQuality ? coordinates : simplifyRadialDistance(coordinates, tolerance: squareTolerance)
+        
+        simplifiedCoordinates = simplifyDouglasPeucker(simplifiedCoordinates, tolerance: squareTolerance)
+
+        return LineString(simplifiedCoordinates)
```
1ec5, line 326: It would be convenient to also offer a mutable `simplify(tolerance:highestQuality:)` method that operates in place, to avoid the copying on this line, though such a method could be tail work.

Sources/Turf/Geometries/LineString.swift
```diff
@@ -215,4 +215,114 @@ extension LineString {
...
+
+        return dx * dx + dy * dy
+    }
+
+    private func simplifyDouglasPeuckerStep(_ coordinates: [CLLocationCoordinate2D], first: Int, last: Int, tolerance: Double, simplified: inout [CLLocationCoordinate2D]) {
+
+        var maxSquareDistance = tolerance
+        var index = 0
+
+        for i in first + 1 ..< last {
```
1ec5, line 277: This algorithm was ported somewhat literally from JavaScript, so it doesn’t make use of functional programming that would insulate us from potential off-by-one errors and the like. To give us more confidence, can you port some of the unit tests from JavaScript to Swift?

Turf.podspec
```diff
@@ -3,7 +3,7 @@ Pod::Spec.new do |s|
   # ―――  Spec Metadata  ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――― #
 
   s.name = "Turf"
-  s.version = "1.1.0"
+  s.version = "1.1.1"
```
1ec5, line 6: Most likely we’ll be releasing this functionality in v1.2.0, but we would take care of the version bump in a separate PR anyways. There’s likely to be a prerelease before the final v1.2.0 release.

README.md
```diff
@@ -92,6 +92,7 @@ Turf.js | Turf-swift
 — | `CLLocationDirection.difference(from:)`
 — | `CLLocationDirection.wrap(min:max:)`
 [turf-polygon-to-line](https://github.com/Turfjs/turf/tree/master/packages/turf-polygon-to-line/) | `LineString(_:)`<br>`MultiLineString(_:)`<br>`FeatureCollection(_:)`
+[turf-simplify](https://github.com/Turfjs/turf/tree/master/packages/turf-simplify) | `simplified(tolerance:highestQuality:)`
```
1ec5, line 95:
```suggestion
[turf-simplify](https://github.com/Turfjs/turf/tree/master/packages/turf-simplify) | `LineString.simplified(tolerance:highestQuality:)`
```

1ec5 (commented):
I’ve lost count of how many times someone has come asking for a Douglas-Peucker simplification method on Apple platforms, coming from Android-land where the Mapbox Java SDK already provides one. Can’t wait to make use of it.

In general, let’s avoid unlabeled arguments unless an unlabeled first argument allows the method name to read more like a sentence. Porting the unit tests over will give us more confidence in the stability and correctness of this port.

Sources/Turf/Geometries/LineString.swift
avi-c: Fixed.

Sources/Turf/Geometries/LineString.swift
avi-c: Tried out a new name that I think makes sense. Also moved the function to be public since it might be generally useful.

Sources/Turf/Geometries/LineString.swift
avi-c: Corrected.

Sources/Turf/Geometries/LineString.swift
avi-c: Added the mutable version as well.

Sources/Turf/Geometries/LineString.swift
avi-c: I brought over the test for LineString from the JS version.
I am curious what changes you'd suggest in the code though.

Turf.podspec
avi-c: Reverted this

README.md
avi-c: Corrected.

Sources/Turf/Geometries/LineString.swift
1ec5: Thanks for bringing over the tests. In my opinion, it’s OK to leave it as a straight port of the JavaScript version if we haven’t gotten to understand the code that deeply. The tests at least give us confidence that we didn’t make an off-by-one error when porting the functionality, especially since the range operators don’t have an equivalent in JavaScript. At some point, we might want to revisit this code to make it more idiomatic and functional, with a possible efficiency benefit. Then the unit tests will also protect us against regressions.

Sources/Turf/Geometries/LineString.swift
```diff
@@ -215,4 +215,132 @@ extension LineString {
         
         return closestCoordinate
     }
+
+    private func squareDistance(from: CLLocationCoordinate2D, to: CLLocationCoordinate2D) -> Double {
```
1ec5, line 219: This is fine for a private method, but FYI it’s more idiomatic to use labels for prepositions, like this:

```swift
private func squareDistance(from origin: CLLocationCoordinate2D, to destination: CLLocationCoordinate2D) -> Double {
```

Sources/Turf/Geometries/LineString.swift
```diff
@@ -215,4 +215,132 @@ extension LineString {
...
+        let dx = from.longitude - to.longitude
+        let dy = from.latitude - to.latitude
+        return dx * dx + dy * dy
+    }
+
+    public func simplified(radialTolerance: Double) -> LineString {
```
1ec5, line 225: This public method needs documentation. Or does it need to be public?

Sources/Turf/Geometries/LineString.swift
```diff
@@ -215,4 +215,132 @@ extension LineString {
...
+    ///
+    /// Ported from https://github.com/Turfjs/turf/blob/master/packages/turf-simplify/lib/simplify.js
+    public mutating func simplified(tolerance: Double = 1.0, highestQuality: Bool = false) {
+        guard coordinates.count > 2 else { return }
+
+        let squareTolerance = tolerance * tolerance
+
+        let simplifiedCoordinates = highestQuality ? coordinates : self.simplified(radialTolerance: squareTolerance).coordinates
+
+        coordinates = simplifyDouglasPeucker(simplifiedCoordinates, tolerance: squareTolerance)
```
1ec5, line 344: This still ends up copying the LineString before mutating the original, because `simplified(radialTolerance:)` creates a new LineString. Ideally, there should be a mutating method `simplified(tolerance:highestQuality:)` that operates directly on the current coordinate array, plus a non-mutating method `simplify(tolerance:highestQuality:)` method that copies self and calls `simplified(tolerance:highestQuality:)` on it.

Sources/Turf/Info.plist
```diff
@@ -15,9 +15,9 @@
 	<key>CFBundlePackageType</key>
 	<string>FMWK</string>
 	<key>CFBundleShortVersionString</key>
-	<string>1.1.0</string>
+	<string>1.1.1</string>
```
1ec5, line 18: There are still some vestiges of the version bump in the Info.plist and .xcodeproj files. I think the upgrade script will paper over this inconsistency, but if you can revert these files that’d be great.

avi-c: Should be fixed now

Sources/Turf/Geometries/LineString.swift
avi-c: Should be fixed. I also did add mutating and non-mutating variants.

Sources/Turf/Geometries/LineString.swift
avi-c: Made it private

Sources/Turf/Geometries/LineString.swift
avi-c: Updated since your format is better.

Sources/Turf/Geometries/LineString.swift
```diff
@@ -215,4 +215,136 @@ extension LineString {
...
+        }
+
+        if prevCoordinate != coordinate {
+            newCoordinates.append(coordinate)
+        }
+
+        coordinates = newCoordinates
+    }
+
+    private func simplify(radialTolerance: Double) -> LineString {
```
1ec5, line 248: This method is unused.

1ec5 (approved):

Issue #102, opened 2020-06-05
BoundingBox mixes up its corners


1ec5:
The `BoundingBox` struct encodes and decodes northwest and southeast coordinates, in that order, whereas the GeoJSON specification [expects](https://tools.ietf.org/html/rfc7946#section-5.1) southwest and northeast coordinates:

https://github.com/mapbox/turf-swift/blob/bd5227e7a4684a6757a55e2bce32b683a6be78ea/Sources/Turf/BoundingBox.swift#L48-L49

If an application decodes GeoJSON, the in-memory representation is incorrect. If an application creates a `BoundingBox` programmatically and encodes it to GeoJSON, the GeoJSON is incorrect.

`BoundingBox` was probably copied from MapboxDirections’ `CoordinateBounds` type, which was fixed in mapbox/mapbox-directions-swift#348.

/cc @mapbox/navigation-ios @frederoni

PR #132, opened 2021-02-05
Rewrite Core Location shim for Linux

### LINKED ISSUE #119: Core Location compatibility shim on Linux uses CL class prefix
This library publicly exposes a compatibility shim for common Core Location types on Linux. The shim works by aliasing primitive types and defining structs from scratch:

https://github.com/mapbox/turf-swift/blob/19d184102927cfc6dba49647850c6fd3e1f5f336/Sources/Turf/CoreLocation.swift#L2-L12

Unfortunately, if an application imports both Turf and another library that implements a similar shim, there could be conflicts that would at least result in client code having to namespace types or import the library more selectively. For example, mapbox/mapbox-directions-swift#488 adds Linux support to a library that imports both Turf and Polyline. Since Turf’s compatibility shim defines types under the original Core Location names, the Linux version of the codebase has to import Turf, even for files that don’t otherwise contain anything related to geometry operations.

In raphaelmor/Polyline#55, I settled on a slightly different approach: standardize the library on a parallel set of slightly differently named types. Primitives continue to be aliased to Core Location types. On Apple platforms, structs are also aliased to the Core Location types, but on Linux, they’re are defined from scratch. This is similar to the Turf approach, but since it doesn’t claim the `CL` class prefix, the developer can explicitly opt into the compatibility shim and choose which library they want to get it from in a given file.

Migrating to the Polyline approach would be a backwards-incompatible change.

/cc @frederoni @Udumft

1ec5:
Rewrote the Core Location compatibility shim for Linux to define non-CL-prefixed types regardless of platform. For example, cross-platform code uses `LocationCoordinate2D` instead of `CLLocationCoordinate2D`, but Apple-platform code can continue to use `CLLocationCoordinate2D`.

Along the way, this PR also renames `BoundingBox(_:_:)` to `BoundingBox(southWest:northEast:)` to avoid mistakes similar to #102 in client code.

Fixes #119.

/cc @mapbox/navigation-ios @macdrevx

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,16 +74,16 @@ Turf.js | Turf-swift
 ----|----
 [turf-along](https://github.com/Turfjs/turf/tree/master/packages/turf-along/) | `LineString.coordinateFromStart(distance:)`
 [turf-area](https://github.com/Turfjs/turf/blob/master/packages/turf-area/) | `Polygon.area`
-[turf-bearing](https://turfjs.org/docs/#bearing) | `CLLocationCoordinate2D.direction(to:)`<br>`RadianCoordinate2D.direction(to:)`
+[turf-bearing](https://turfjs.org/docs/#bearing) | `CLLocationCoordinate2D.direction(to:)`<br>`LocationCoordinate2D.direction(to:)` on Linux<br>`RadianCoordinate2D.direction(to:)`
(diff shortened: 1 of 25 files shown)


1ec5:
This PR looks frighteningly large, but most of it is just removing the `CL` prefix from every call site. The only interesting changes are in CoreLocation.swift and BoundingBox.swift.

macdrevx (approved):

PR #133, opened 2021-02-09
Enable library evolution and prepare v2.0.0-alpha.2

macdrevx:
- Enables library evolution in the Xcode project
- Enables library evolution in the podspec
- Updates version numbers for v2.0.0-alpha.2 release
- Configures build settings for generating XCFrameworks
- Adds a script to generate an XCFramework
- Configures the Package.swift to use binary distribution


Package.swift
```diff
@@ -1,28 +1,21 @@
...
 
 import PackageDescription
+import Foundation
+
+let checksum = "5256bc0056fb35d19f3d4b5a59d9d22e6a2d57e98664e46de656764a71c7d5e0"
+let url = "https://github.com/mapbox/turf-swift/releases/download/v2.0.0-alpha.2/Turf.xcframework.zip"
 
 let package = Package(
     name: "Turf",
+    platforms: [.iOS(.v10), .macOS(.v10_12)],
```
1ec5, line 12: Is this binary being built for macOS? If not, remove that item from the array to make incompatibility clearer to anyone who accidentally ends up with this prerelease.

Turf and MapboxDirections both support macOS, tvOS, watchOS, and Linux in addition to iOS. Normally we would not be able to regress support for these platforms, but we’ll do so for this alpha only.

scripts/create-xcframework.sh
```diff
@@ -0,0 +1,49 @@
...
+  -framework "$TURF_IOS_SIMULATOR_XCARCHIVE/$TURF_FRAMEWORK_PATH" \
+  -framework "$TURF_IOS_XCARCHIVE/$TURF_FRAMEWORK_PATH" \
+  -framework "$TURF_MAC_CATALYST_XCARCHIVE/$TURF_FRAMEWORK_PATH" \
+  -framework "$TURF_MACOS_XCARCHIVE/$TURF_FRAMEWORK_PATH" \
+  -output "$TURF_BUILD_DIR/Turf.xcframework"
+
+pushd "$TURF_BUILD_DIR"
+zip -r "Turf.xcframework.zip" "Turf.xcframework"
+popd
+
```
neelmistry94, line 48: is this something we'd want to make part of CI. Or leave it as is since the xcframework is a stop gap?

1ec5 (approved):
The build setting change and limited platform support would normally be unmergeable. However, this PR targets a branch from which we’ll tag an alpha that’s specifically intended as a stopgap to unblock mapbox/mapbox-navigation-ios#2629 and mapbox/mapbox-navigation-ios#2794 for a prerelease. Once the map SDK migrates to source-based distribution, this branch and the corresponding one in mapbox-directions-swift will go away and we’ll continue to build as before.

It looks like the rest of the changes are consistent with running scripts/update-version.sh.

1ec5:
[Linux builds are failing](https://travis-ci.com/github/mapbox/turf-swift/builds/216569534) because they’re still on Swift 5, whereas this PR now requires Swift 5.3:

```
$ swift build

/home/travis/build/mapbox/turf-swift: error: package at '/home/travis/build/mapbox/turf-swift' is using Swift tools version 5.3.0 but the installed version is 5.0.0

The command "swift build" exited with 1.
```

Package.swift
macdrevx: The binary was built for iOS and macOS.

scripts/create-xcframework.sh
macdrevx: Let's leave it as-is since it's a stop gap solution. I included it for transparency so that anyone can see how the XCFramework was generated.

PR #134, opened 2021-02-22
Enable library evolution (redux)

1ec5:
This is a whittled-down version of #133 that enables the `BUILD_FOR_DISTRIBUTION` setting only in the Xcode project for use with Carthage and continues to distribute the library as source code, preserving support for operating systems other than iOS and macOS.

/cc @macdrevx @neelmistry94 @julianrex @chezzdev


Turf.podspec
```diff
@@ -3,7 +3,7 @@ Pod::Spec.new do |s|
   # ―――  Spec Metadata  ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――― #
 
   s.name = "Turf"
-  s.version = "1.2.0"
+  s.version = "2.0.0-alpha.2"
```
1ec5, line 6: I’ll take care of increasing the version to v2.0.0-alpha.3 in a separate PR for clarity.

macdrevx (approved):

PR #137, opened 2021-03-04
Add polygon smooth

okcoker:
Went through the trouble of porting this and realized I didn't even need it. Hopefully my loss is someone's gain 🤦


Udumft (approved):
LGTM. Thank you for contributing!

PR #138, opened 2021-03-04
Add Polygon simplify

okcoker:
Added Polygon simplify.

The expected output numbers had to be altered with more precision from what you see [here](https://github.com/Turfjs/turf/blob/402716a29f6ae16bf3d0220e213e5380cc5a50c4/packages/turf-simplify/test/out/simple-polygon.geojson#L10-L13)

I noticed this was done with the `LineString` tests as well.

### DIFF
--- a/Sources/Turf/Geometries/Polygon.swift
+++ b/Sources/Turf/Geometries/Polygon.swift
@@ -140,4 +140,163 @@ extension Polygon {
             tempOutput[i] = ring + [ring[0]]
         })
     }
+
+    /// Returns a copy of the Polygon with the Ramer–Douglas–Peucker algorithm applied to it.
+    ///
+    /// tolerance:  Controls the level of simplification by specifying the maximum allowed distance between the original line point
+    /// and the simplified point. Higher tolerance values results in higher simplification.
+    ///
+    /// highestQuality: Excludes distance-based preprocessing step which leads to highest quality simplification. High quality simplification runs considerably slower so consider how much precision is needed in your application.
+    ///
+    /// Ported from https://github.com/Turfjs/turf/blob/master/packages/turf-simplify/lib/simplify.js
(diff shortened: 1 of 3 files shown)


Sources/Turf/Geometries/Polygon.swift
```diff
@@ -76,4 +76,163 @@ extension Polygon {
...
+
+    /**
+     * Returns true if ring has at least 3 coordinates and its first coordinate is the same as its last
+     *
+     * @private
+     * @param {Array<number>} ring coordinates to be checked
+     * @returns {boolean} true if valid
+     */
+    private func checkValidity(ring: [LocationCoordinate2D]) -> Bool {
+        if ring.count < 3 { return false }
```
Udumft, line 230: Usually, `guard` is used for such checks.

Sources/Turf/Geometries/Polygon.swift
```diff
@@ -76,4 +76,163 @@ extension Polygon {
...
+
+        let lastPoint = coordinates.count - 1
+        var result = [coordinates[0]]
+        simplifyDouglasPeuckerStep(coordinates, first: 0, last: lastPoint, tolerance: tolerance, simplified: &result)
+        result.append(coordinates[lastPoint])
+        return result
+    }
+
+    /**
+     * Returns true if ring has at least 3 coordinates and its first coordinate is the same as its last
```
Udumft, line 223: This description does not match the logic of the function: if a ring has 3 coordinates, and it's first and last coordinates are equal - it will return `false`. Thus, the function checks first and last coordinates only for 3-points rings.

Udumft (commented):
A few comments:

Sources/Turf/Geometries/Polygon.swift
okcoker: Ah yes the original comment set me up for failure. Just pushed an update.

Udumft (approved):
LGTM, thanks for contributing!

PR #140, opened 2021-03-19
Make bezier feature on LineString public.

Woildan:
Features in the LineString.swift file are all public except for the bezier function, which we need in our projet.
This PR fixes this.


1ec5 (approved):
Thanks!

PR #143, opened 2021-04-09
Use Measurement API instead of RadianDirection typealias

### LINKED ISSUE #38: Replace RadianDirection with UnitAngle
This library represents radians as a `RadianDirection` type alias. Instead, we can use the built-in `Measurement` struct in conjunction with `UnitAngle` to convert between degrees and radians.

/cc @frederoni @captainbarbosa

azarovalex:
Fixes #38.
New Measurement API is available from iOS 10 and allows to specify concrete units for all measurements.
Previously we expected a Double value in radians, however, it was possible for the user to pass Double value in degrees.
This PR ensures that the user will pass only values that are measured in radians.

### DIFF
--- a/Sources/Turf/CoreLocation.swift
+++ b/Sources/Turf/CoreLocation.swift
@@ -188,12 +188,13 @@ extension LocationCoordinate2D: Equatable {
     
     /// Returns the direction from the receiver to the given coordinate.
     public func direction(to coordinate: LocationCoordinate2D) -> LocationDirection {
-        return RadianCoordinate2D(self).direction(to: RadianCoordinate2D(coordinate)).toDegrees()
+        return RadianCoordinate2D(self).direction(to: RadianCoordinate2D(coordinate)).converted(to: .degrees).value
     }
     
     /// Returns a coordinate a certain Haversine distance away in the given direction.
     public func coordinate(at distance: LocationDistance, facing direction: LocationDirection) -> LocationCoordinate2D {
-        let radianCoordinate = RadianCoordinate2D(self).coordinate(at: distance / metersPerRadian, facing: direction.toRadians())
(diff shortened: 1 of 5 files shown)


Sources/Turf/CoreLocation.swift
```diff
@@ -188,12 +188,13 @@ extension LocationCoordinate2D: Equatable {
...
     /// Returns the direction from the receiver to the given coordinate.
     public func direction(to coordinate: LocationCoordinate2D) -> LocationDirection {
-        return RadianCoordinate2D(self).direction(to: RadianCoordinate2D(coordinate)).toDegrees()
+        return RadianCoordinate2D(self).direction(to: RadianCoordinate2D(coordinate)).converted(to: .degrees).value
     }
     
     /// Returns a coordinate a certain Haversine distance away in the given direction.
     public func coordinate(at distance: LocationDistance, facing direction: LocationDirection) -> LocationCoordinate2D {
-        let radianCoordinate = RadianCoordinate2D(self).coordinate(at: distance / metersPerRadian, facing: direction.toRadians())
+        let radiansDirection = Measurement(value: direction, unit: UnitAngle.degrees)
```
1ec5, line 196: It wouldn’t be strictly necessary, but it could be nice to offer a `LocationCoordinate2D.coordinate(at:facing:)` that takes a `Measurement<UnitAngle>` just like the one on `RadianCoordinate2D`. Then this method becomes a lightweight wrapper around it that just converts the `LocationDirection` to a degree measurement. But that’d only be for completeness’ sake.

1ec5 (approved):
Looks great!

Sources/Turf/CoreLocation.swift
azarovalex: done!

PR #144, opened 2021-04-16
Vendor swiftenv installation script

1ec5:
This PR copies the swiftenv installation script from https://swiftenv.fuller.li/en/latest/install.sh, since the script is stable and quite simple. It is unfortunately still necessary because the Travis image still has the Swift 4._x_ toolchain installed by default.

/cc @mapbox/navigation-ios

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,8 @@ language: generic
 sudo: required
 dist: trusty
 env:
-  - SWIFT_VERSION=5.0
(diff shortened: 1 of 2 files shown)


1ec5:
Travis builds are [failing](https://travis-ci.com/github/mapbox/turf-swift/builds/223318684) because apparently the vendored script doesn’t lay down `swift` in the same place that the remote script did?

> The program 'swift' is currently not installed. To run 'swift' please ask your administrator to install the package 'python-swiftclient'

avi-c (approved):

PR #147, opened 2021-05-28
v2.0.0-alpha.4

chezzdev:


Turf.podspec
```diff
@@ -3,7 +3,7 @@ Pod::Spec.new do |s|
   # ―――  Spec Metadata  ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――― #
 
   s.name = "Turf"
-  s.version = "2.0.0-alpha.3"
+  s.version = "2.0.0-alpha.4"
```
1ec5, line 6: Let’s make it v2.0.0-beta.1, for consistency with MapboxDirections and MapboxCoreNavigation, which have long been in beta.

1ec5 (approved):
It should be a beta, but I’ll approve so you don’t need to get another round of reviews after making that change.

Issue #145, opened 2021-04-23
Is there a way to get the center of the polygon?


bagusflyer:
Is there a way to get the center of the polygon by using `turf-swift`?

Udumft:
Hello, @bagusflyer! Unfortunately there is no API method to do that. But I think it should be as simple as calculating "average" among all of Polygon's coordinates.

1ec5:
> But I think it should be as simple as calculating "average" among all of Polygon's coordinates.

That’s one kind of center, but it only yields reasonable results for convex polygons.

Turf.js has several center- and centroid-related functions that we could port to Swift:

* [`center`](http://turfjs.org/docs/#center) is similar to the “average” described above.
* [`centerOfMass`](http://turfjs.org/docs/#centerOfMass)
* [`centroid`](http://turfjs.org/docs/#centroid)

The iOS map SDK (until v10) has an `MGLPolyline.coordinate` property that’s very useful for this purpose. Under the hood, it relies on [Polylabel](https://github.com/mapbox/polylabel/), a C++ implementation of the pole-of-inaccessibility algorithm (equivalent to [`ST_MaximumInscribedCircle`](https://postgis.net/docs/ST_MaximumInscribedCircle.html) in PostGIS). There are other centroid algorithms, but the pole of inaccessibility algorithm yields very good results overall.

Issue #118, opened 2020-10-31
Turf::intersection does not observe edge cases


westnordost:
I think this function does not observe the following edge cases:

- both lines are colinear (denominator == 0): the line segments could still intersect - the intersection would be another line segment
- what about the 180th meridian? what about line segments crossing the north and south pole?
- the current algorithm does not seem to count the start end end point as part of the line segment. This is a matter of definition, but usually at least the start point is counted as part of the line (i.e. the line segment is then closed-open)

1ec5:
https://github.com/mapbox/turf-swift/blob/19d184102927cfc6dba49647850c6fd3e1f5f336/Sources/Turf/Turf.swift#L15-L16

turf-point-on-line has since been replaced by [turf-nearest-point-on-line](https://github.com/Turfjs/turf/blob/23d5cb91d77e0c1e2e903a2252f525797f1d0d09/packages/turf-nearest-point-on-line/index.ts), which calls out to [turf-line-intersect](https://github.com/Turfjs/turf/blob/23d5cb91d77e0c1e2e903a2252f525797f1d0d09/packages/turf-line-intersect/index.ts), which at a glance seems to handle two of these issues.

> what about the 180th meridian? what about line segments crossing the north and south pole?

This is a good point; geometries that straddle the antimeridian probably aren’t handled well by other methods in this library either. GeoJSON technically represents such geometries as multi-geometries, but many clients including the Mapbox Maps SDK work around this limitation by allowing longitudes to extend beyond ±180°.

Issue #26, opened 2017-12-15
Haversine distance should use spherical Earth radius


1ec5:
This library currently defines the radius of the Earth as 6&nbsp;373&nbsp;000&nbsp;m for the purpose of converting between meters and radians in the Haversine formula:

https://github.com/mapbox/turf-swift/blob/40c7d772c46837aa9e5ceb60d22d8ad97a4be708/Turf/Turf.swift#L7

Apparently this value came from [an old version of turf-distance](https://github.com/Turfjs/turf-distance/blob/cea4c6cbec582bff8880193a3c2cda38ae731967/index.js#L78) that I originally ported to an internal Swift application, before it got copied into another internal Swift application, then copied into the navigation SDK, then moved here. It’s close to the value of 6&nbsp;372&nbsp;797.560&nbsp;856 that [Osmium](https://github.com/Project-OSRM/osrm-backend/blob/bb1f4a025a3cefd3598a38b9d3e55485d1080ec5/third_party/libosmium/include/osmium/geom/haversine.hpp#L58-L59) describes as “Earth’s quadratic mean radius for WGS84”.

These days, Turf.js uses [a spherical approximation](https://github.com/Turfjs/turf/blob/a94151418cb969868fdb42955a19a133512da0fd/packages/turf-helpers/index.js#L4) of 6&nbsp;671&nbsp;008.8&nbsp;m for its radian-to-meter conversion and Haversine formula. The Haversine formula was always meant to be used with a spherical meters-per-radian value. We should probably change this value to match Turf.js. 🙀

/ref Turfjs/turf#978 Turfjs/turf#1012 Turfjs/turf#1176 https://github.com/mapbox/turf-swift/pull/21#discussion_r157246151
/cc @frederoni @bsudekum @captainbarbosa

bsudekum:
Do you think this could have any impact on https://github.com/mapbox/mapbox-navigation-ios/issues/901?

1ec5:
> Do you think this could have any impact on https://github.com/mapbox/mapbox-navigation-ios/issues/901?

Yes, that issue could well be caused by a discrepancy between OSRM and Turf that builds up as per-segment and per-step distances are summed together. OSRM uses an Earth radius of 6&nbsp;372&nbsp;797.560&nbsp;856 when calculating distances:

[...]

PR #148, opened 2021-07-19
Add Polygon centroid and center of mass

nighthawk:
This ports Turf's [`centerOfMass`](http://turfjs.org/docs/#centerOfMass) and [`centroid`](http://turfjs.org/docs/#centroid) to Swift, at least for the individual `Polygon` case. I've followed the same implementation and test cases as from the JS repo. 

Note: This could be extended to support the other geometries that Turf.js supports, though this should then also add support for convex hull calculations.

This addresses part of #145.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,8 @@ Turf.js | Turf-swift
 [turf-bearing](https://turfjs.org/docs/#bearing) | `CLLocationCoordinate2D.direction(to:)`<br>`LocationCoordinate2D.direction(to:)` on Linux<br>`RadianCoordinate2D.direction(to:)`
 [turf-bezier-spline](https://github.com/Turfjs/turf/tree/master/packages/turf-bezier-spline/) | `LineString.bezier(resolution:sharpness:)`
 [turf-boolean-point-in-polygon](https://github.com/Turfjs/turf/tree/master/packages/turf-boolean-point-in-polygon) | `Polygon.contains(_:ignoreBoundary:)`
+[turf-center-of-mass](http://turfjs.org/docs/#centerOfMass) | `Polygon.centerOfMass()` |
+[turf-centroid](http://turfjs.org/docs/#centroid) | `Polygon.centroid()` |
 [turf-circle](https://turfjs.org/docs/#circle) | `Polygon(center:radius:vertices:)` |
 [turf-destination](https://github.com/Turfjs/turf/tree/master/packages/turf-destination/) | `CLLocationCoordinate2D.coordinate(at:facing:)`<br>`LocationCoordinate2D.coordinate(at:facing:)` on Linux<br>`RadianCoordinate2D.coordinate(at:facing:)`
 [turf-distance](https://github.com/Turfjs/turf/tree/master/packages/turf-distance/) | `CLLocationCoordinate2D.distance(to:)`<br>`LocationCoordinate2D.distance(to:)` on Linux<br>`RadianCoordinate2D.distance(to:)`
--- a/README.md
+++ b/README.md
@@ -75,8 +75,9 @@ Turf.js | Turf-swift
 [turf-bearing](https://turfjs.org/docs/#bearing) | `CLLocationCoordinate2D.direction(to:)`<br>`LocationCoordinate2D.direction(to:)` on Linux<br>`RadianCoordinate2D.direction(to:)`
 [turf-bezier-spline](https://github.com/Turfjs/turf/tree/master/packages/turf-bezier-spline/) | `LineString.bezier(resolution:sharpness:)`
 [turf-boolean-point-in-polygon](https://github.com/Turfjs/turf/tree/master/packages/turf-boolean-point-in-polygon) | `Polygon.contains(_:ignoreBoundary:)`
-[turf-center-of-mass](http://turfjs.org/docs/#centerOfMass) | `Polygon.centerOfMass()` |
-[turf-centroid](http://turfjs.org/docs/#centroid) | `Polygon.centroid()` |
+[turf-center](http://turfjs.org/docs/#center) | `Polygon.center`
+[turf-center-of-mass](http://turfjs.org/docs/#centerOfMass) | `Polygon.centerOfMass`
+[turf-centroid](http://turfjs.org/docs/#centroid) | `Polygon.centroid`
 [turf-circle](https://turfjs.org/docs/#circle) | `Polygon(center:radius:vertices:)` |
 [turf-destination](https://github.com/Turfjs/turf/tree/master/packages/turf-destination/) | `CLLocationCoordinate2D.coordinate(at:facing:)`<br>`LocationCoordinate2D.coordinate(at:facing:)` on Linux<br>`RadianCoordinate2D.coordinate(at:facing:)`
 [turf-distance](https://github.com/Turfjs/turf/tree/master/packages/turf-distance/) | `CLLocationCoordinate2D.distance(to:)`<br>`LocationCoordinate2D.distance(to:)` on Linux<br>`RadianCoordinate2D.distance(to:)`
--- a/Sources/Turf/CoreLocation.swift
+++ b/Sources/Turf/CoreLocation.swift
@@ -54,12 +54,12 @@ public struct LocationCoordinate2D {
     /**
      The latitude in degrees.
      */
-    public internal(set) var latitude: LocationDegrees
+    public var latitude: LocationDegrees
     
     /**
      The longitude in degrees.
      */
-    public internal(set) var longitude: LocationDegrees
+    public var longitude: LocationDegrees
     
     /**
      Creates a degree-based geographic coordinate.
--- a/Sources/Turf/CoreLocation.swift
+++ b/Sources/Turf/CoreLocation.swift
@@ -54,12 +54,12 @@ public struct LocationCoordinate2D {
     /**
      The latitude in degrees.
      */
-    public let latitude: LocationDegrees
+    public internal(set) var latitude: LocationDegrees
     
     /**
      The longitude in degrees.
      */
-    public let longitude: LocationDegrees
+    public internal(set) var longitude: LocationDegrees
     
     /**
      Creates a degree-based geographic coordinate.
--- a/Sources/Turf/CoreLocation.swift
+++ b/Sources/Turf/CoreLocation.swift
@@ -71,6 +71,18 @@ public struct LocationCoordinate2D {
 }
 #endif
 
+extension LocationCoordinate2D {
+    /**
+        Returns a normalized coordinate, wrapped to -180 and 180 degrees latitude
+     */
+    var normalized: LocationCoordinate2D {
+        return .init(
+            latitude: latitude,
+            longitude: longitude.wrap(min: -180, max: 180)
+        )
+    }
+}
+
 extension LocationDirection {
     /**
      Returns a normalized number given min and max bounds.
--- a/Sources/Turf/Geometries/Polygon.swift
+++ b/Sources/Turf/Geometries/Polygon.swift
@@ -321,7 +321,7 @@ extension Polygon {
                 acc.latitude += next.latitude
                 acc.longitude += next.longitude
             }
-        return .init(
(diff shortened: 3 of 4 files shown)


Sources/Turf/CoreLocation.swift
```diff
@@ -54,12 +54,12 @@ public struct LocationCoordinate2D {
     /**
      The latitude in degrees.
      */
-    public let latitude: LocationDegrees
+    public internal(set) var latitude: LocationDegrees
```
1ec5, line 57: It looks like CLLocationCoordinate2D declares these properties as variable even for their getters, so ideally this Linux-specific declaration would be consistent with it.

Sources/Turf/Geometries/Polygon.swift
```diff
@@ -297,4 +297,55 @@ extension Polygon {
             ring[2].longitude == ring[0].longitude
         )
     }
-}
+
+
+    /// Calculates the centroid using the mean of all vertices.
+    /// This lessens the effect of small islands and artifacts when calculating the centroid of a set of polygons.
+    public func centroid() -> LocationCoordinate2D? {
```
1ec5, line 304: Polygon already has `area` as a computed property rather than a method, so having `centroid` and `centerOfMass` be computed properties too would be more elegant. There’s probably a point at which a parameterless operation is expensive enough that the method syntax is more appropriate, but `centroid` ends up being a single reduction, which is similar to `area`.

Sources/Turf/Geometries/Polygon.swift
```diff
@@ -297,4 +297,55 @@ extension Polygon {
...
+        // We take any point to translate all the points around 0
+        guard let center = centroid() else { return nil }
+        let coordinates = outerRing.coordinates
+        let neutralized = coordinates.map {
+            LocationCoordinate2D(latitude: $0.latitude - center.latitude, longitude: $0.longitude - center.longitude)
+        }
+        
+        var signedArea: Double = 0
+        var sum = LocationCoordinate2D(latitude: 0, longitude: 0)
+        let zipped = zip(neutralized.prefix(upTo: neutralized.count - 1), neutralized.suffix(from: 1))
```
1ec5, line 329: Good use of `zip(_:_:)`!

Sources/Turf/Geometries/Polygon.swift
```diff
@@ -297,4 +297,55 @@ extension Polygon {
...
+
+
+    /// Calculates the centroid using the mean of all vertices.
+    /// This lessens the effect of small islands and artifacts when calculating the centroid of a set of polygons.
+    public func centroid() -> LocationCoordinate2D? {
+        let coordinates = outerRing.coordinates.dropLast()
+        let summed = coordinates
+            .reduce(into: LocationCoordinate2D(latitude: 0, longitude: 0)) { acc, next in
+                acc.latitude += next.latitude
+                acc.longitude += next.longitude
```
1ec5, line 309: Should we be concerned about features that span the antimeridian? Technically, GeoJSON doesn’t allow a single feature to span the antimeridian, but it’s common (at least among Mapbox products) to represent it as a single feature with coordinates that extend beyond ±180°. If this implementation doesn’t handle that case correctly, that’s OK; we can open a separate issue for it.

Sources/Turf/Geometries/Polygon.swift
```diff
@@ -297,4 +297,55 @@ extension Polygon {
             ring[2].longitude == ring[0].longitude
         )
     }
-}
+
+
+    /// Calculates the centroid using the mean of all vertices.
+    /// This lessens the effect of small islands and artifacts when calculating the centroid of a set of polygons.
```
1ec5, line 303: I see you’ve added permalinks to the JavaScript tests that you ported; can you add a similar permalink to the implementation that you ported too? That way we can easily update the implementation if anything changes in the JavaScript version.

Sources/Turf/Geometries/Polygon.swift
```diff
@@ -297,4 +297,55 @@ extension Polygon {
...
+    public func centroid() -> LocationCoordinate2D? {
+        let coordinates = outerRing.coordinates.dropLast()
+        let summed = coordinates
+            .reduce(into: LocationCoordinate2D(latitude: 0, longitude: 0)) { acc, next in
```
1ec5, line 307: The JavaScript implementation [relies on the feature’s bounding box](https://github.com/Turfjs/turf/blob/ae066ae320a5e6a95960f2ed608d79d52b7b62d8/packages/turf-center/index.ts#L40). I’d imagine that this implementation is a bit more efficient than the one in `BoundingBox(from:)`. Perhaps we could replace it outright?

Sources/Turf/Geometries/Polygon.swift
```diff
@@ -297,4 +297,55 @@ extension Polygon {
             ring[2].longitude == ring[0].longitude
         )
     }
-}
+
+
+    /// Calculates the centroid using the mean of all vertices.
+    /// This lessens the effect of small islands and artifacts when calculating the centroid of a set of polygons.
+    public func centroid() -> LocationCoordinate2D? {
+        let coordinates = outerRing.coordinates.dropLast()
```
1ec5, line 305: In the future, once similar calculations are implemented for other geometry types, Geometry itself could have `centroid` and `centerOfMass` properties that pass through to type-specific implementations like this.

1ec5 (commented):
Thank you for porting the JavaScript implementation of these methods – I’m sure a lot of developers will find them quite handy. I haven’t gone line by line comparing the port to the original JavaScript, but let me know if you’d like a second pair of eyes on that.

Sources/Turf/Geometries/Polygon.swift
nighthawk: They are different things. The JS implementation that you linked is for `center`. I haven't ported that one, but rather `centroid` (as `centerOfMass` depends on that). Both have different advantages and disadvantages. The bounding boxes `center` is susceptible to outliers, while `centroid` is susceptive to inbalanced polygons.

See for example:
- center: https://github.com/Turfjs/turf/blob/master/packages/turf-center/test/out/imbalanced-polygon.geojson
- centroid: https://github.com/Turfjs/turf/blob/master/packages/turf-centroid/test/out/imbalanced-polygon.geojson

I can add `center`, too, as it's pretty simple.

Sources/Turf/CoreLocation.swift
nighthawk: Good point. Done.

Sources/Turf/Geometries/Polygon.swift
nighthawk: Makes sense. I've renamed the two for consistency.

Sources/Turf/Geometries/Polygon.swift
nighthawk: Good point. I don't think it handles that properly. It should check if the final longitude extends beyond ±180° and then snap it back within the allowed range. Agree that it might be better to create a separate issue for that. (Turf.js might also need fixing for that.)

Sources/Turf/Geometries/Polygon.swift
nighthawk: Sure thing. Done.

Sources/Turf/Geometries/Polygon.swift
1ec5: The fix would be to wrap the longitude to the ±180° range. Snapping would result in truncating the geometry at the antimeridian, which would be unexpected.

1ec5:
@nighthawk, feel free to push up a revised version when you’re ready, or let us know if you need help. FYI, we’re planning to release v2.0.0 in the coming weeks, but these changes are backwards-compatible, so we can land them in a subsequent minor version if the timing works out better.

Sources/Turf/Geometries/Polygon.swift
nighthawk: Done!

nighthawk:
@1ec5 , no worries. I've pushed a revised version. Please have a look.

A little curiosity that I encountered when implementing `center`: Doing the manual calculation of the latitude and longitude of the bounding boxes center as commited leads to a different result that doing `mid(bbox.southWest, bbox.northEast)`. For the Lyon example it was off by 5 metres. Is that due to the curvature of the Earth? I went with the manual calculation for consistency with Turf.js.

Sources/Turf/Geometries/Polygon.swift
```diff
@@ -298,27 +298,42 @@ extension Polygon {
...
             .reduce(into: LocationCoordinate2D(latitude: 0, longitude: 0)) { acc, next in
                 acc.latitude += next.latitude
                 acc.longitude += next.longitude
             }
-        return .init(
+        return LocationCoordinate2D(
             latitude: summed.latitude / Double(coordinates.count),
             longitude: summed.longitude / Double(coordinates.count)
-        )
+        ).normalized
```
1ec5, line 327: I was hasty in suggesting “wrapping” above: it’s more correct than snapping but maybe not what a developer would expect. If a feature (such as Alaska) straddles the antimeridian, I think this will cause the bounding box to include everything except Alaska at the same latitudes, effectively going the longer way around the world. It could be that your original code worked fine, but a test would give us more confidence. If not, I think it’s OK to restore the original version (which matches Turf.js more closely) and track this edge case separately, as with #118.

README.md
```diff
@@ -75,6 +75,8 @@ Turf.js | Turf-swift
 [turf-bearing](https://turfjs.org/docs/#bearing) | `CLLocationCoordinate2D.direction(to:)`<br>`LocationCoordinate2D.direction(to:)` on Linux<br>`RadianCoordinate2D.direction(to:)`
 [turf-bezier-spline](https://github.com/Turfjs/turf/tree/master/packages/turf-bezier-spline/) | `LineString.bezier(resolution:sharpness:)`
 [turf-boolean-point-in-polygon](https://github.com/Turfjs/turf/tree/master/packages/turf-boolean-point-in-polygon) | `Polygon.contains(_:ignoreBoundary:)`
+[turf-center-of-mass](http://turfjs.org/docs/#centerOfMass) | `Polygon.centerOfMass()` |
+[turf-centroid](http://turfjs.org/docs/#centroid) | `Polygon.centroid()` |
```
1ec5, line 79:
```suggestion
[turf-center](http://turfjs.org/docs/#center) | `Polygon.center`
[turf-center-of-mass](http://turfjs.org/docs/#centerOfMass) | `Polygon.centerOfMass`
[turf-centroid](http://turfjs.org/docs/#centroid) | `Polygon.centroid`
```

Tests/TurfTests/PolygonTests.swift
```diff
@@ -142,6 +142,153 @@ class PolygonTests: XCTestCase {
 
         XCTAssertEqual(expectedDiameter, diameter, accuracy: 0.25)
     }
+    
+    func testPolygonCentre() {
+        // Adopted from https://github.com/Turfjs/turf/blob/3b20c568e5638f680cde39c26b56fbcf034133f2/packages/turf-center/test.js
+        let coordinate = LocationCoordinate2D(latitude: 45.7536760235992, longitude: 4.841880798339844)
```
1ec5, line 148: I suppose the Lyon example would be more offset than expected when using `mid(_:_:)` because of a longstanding issue in which `LocationCoordinate2D.distance(to:)` uses a different approximation of the radius of the Earth than Turf.js does these days: #26.

1ec5 (commented):
Thank you for porting the full complement of centroid-related methods. Just to keep things simple for now, let’s roll back the changes related to the antimeridian and track that issue separately. Otherwise, this PR will be good to go. Thanks!

Sources/Turf/Geometries/Polygon.swift
1ec5: Oh, my mistake, `normalized` is being applied to the final result, not each individual coordinate. So this will do the right thing. Sorry for the confusion.

1ec5 (approved):
Ignore my previous review above, this is actually correct. 😅

PR #149, opened 2021-08-05
Release v2.0.0-rc.1

Udumft:
Release v2.0.0-rc.1


Turf.podspec
```diff
@@ -3,7 +3,7 @@ Pod::Spec.new do |s|
   # ―――  Spec Metadata  ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――― #
 
   s.name = "Turf"
-  s.version = "2.0.0-beta.1"
+  s.version = "2.0.0-rc1"
```
azarovalex, line 6: `rc.1`?

azarovalex (approved):
lgtm, just one minor nit

Turf.podspec
bamx23: IMO it's important as `rc1` and `rc.1` are not the same in terms of semver

Udumft: Yes, we've [discussed](https://mapbox.slack.com/archives/CU0LHNNJY/p1628065685013700) that it is OK to go for RC for Turf since there are no more changes planned.

Udumft: Oh, now I see there's a misstype!

PR #151, opened 2021-09-05
Various bug fixes for LineString.simplify and Polygon.simplify

nighthawk:
This branch fixes these issues introduced in the port of turf-simplify from #138:

- `LineString.simplify` with `highestQuality = false` was broken as it was using `start.coordinate` in one formula when it should have used `coordinate`
- `Polygon.simplify` ignored `highestQuality = false` as it first assigned the `self.coordinates` to a local variable, then simplified the `self.coordinates`, but then passed the previous local variable to the Douglas-Peucker algorithm. That means it always returned results as if `highestQuality = true` was set.
- `Polygon.simplify` forgot to square the tolerance to the Douglas-Peucker algorithm (`LineString.simplify` did it correctly)
- `Polygon.simplify` didn't simplify if *any* of the exterior/interior polygons was a triangle; now it just skips simplification of the triangle but simplifies the others.

I've also refactored the code to remove the code duplication between `LineString.swift` and `Polygon.swift`, and ported the turf-simplify test suite using the in/out directory -- excluding the test cases for MultiLineString and MultiPolygon as they aren't supported yet.

---

TODO:

- [x] Port the full test suite of `turf-simplify` to this repo
- [x] Investigate remaining test failure for `polygon.geojson`


nighthawk:
This is ready for review. It runs fine locally and the tests pass. I can't access the bitrise issue for tvOS, so if someone can point me at what's wrong, I'm happy to investigate.

nighthawk:
Oops. I accidentally deleted my branch. Re-instated as #152.

PR #152, opened 2021-09-05
Various bug fixes for LineString.simplify and Polygon.simplify

nighthawk:
This branch fixes these issues introduced in the port of turf-simplify from #138:

- `LineString.simplify` with `highestQuality = false` was broken as it was using `start.coordinate` in one formula when it should have used `coordinate`
- `Polygon.simplify` ignored `highestQuality = false` as it first assigned the `self.coordinates` to a local variable, then simplified the `self.coordinates`, but then passed the previous local variable to the Douglas-Peucker algorithm. That means it always returned results as if `highestQuality = true` was set.
- `Polygon.simplify` forgot to square the tolerance to the Douglas-Peucker algorithm (`LineString.simplify` did it correctly)
- `Polygon.simplify` didn't simplify if *any* of the exterior/interior polygons was a triangle; now it just skips simplification of the triangle but simplifies the others.

I've also refactored the code to remove the code duplication between `LineString.swift` and `Polygon.swift`, and ported the turf-simplify test suite using the in/out directory -- excluding the test cases for MultiLineString and MultiPolygon as they aren't supported yet.


Tests/TurfTests/TurfTests.swift
```diff
@@ -158,4 +158,69 @@ class TurfTests: XCTestCase {
...
+            return [feature]
+        } else if let featureCollection = self.decodedFeatureCollection {
+            return featureCollection.features
+        } else {
+            return []
+        }
+    }
+}
+
+func XCTAssertEqual(_ expected: [[LocationCoordinate2D]], _ actual: [[LocationCoordinate2D]], accuracy: CLLocationDegrees, _ message: @autoclosure () -> String = "", file: StaticString = #filePath, line: UInt = #line) {
```
1ec5, line 209: Some of the Bitrise builds are failing because `#filePath` requires Swift 5.3, which is in Xcode 12. The readme currently states a minimum requirement of Xcode 9, but this would be a good opportunity to increase the requirement to 12.

Sources/Turf/Simplifier.swift
```diff
@@ -0,0 +1,109 @@
+import Foundation
+
+enum Simplifier {
```
1ec5, line 3: Interesting use of an enumeration as an effective namespace, rather than a struct that’s more commonly used for this purpose but comes with initializers. This pattern might be worth following elsewhere in the codebase.

1ec5 (commented):
Great work! Putting the test fixtures in JSON instead of inline feels a lot cleaner.

1ec5 (approved):
Tests are passing now that I’ve upgraded the tvOS Bitrise application to Xcode 12.5.

1ec5:
6ecf55bc9553d2073e8fbb8754c4e0a27223ea4f updates the readme to require Xcode 12.0.

1ec5:
c120ea5838f1f0dca1f6bcb5f6bb6af87829be3a adds the test fixtures to the Xcode project for Carthage-based builds.

Issue #84, opened 2019-12-18
CLLocationCoordinate2D conformance to Codable should be private


1ec5:
Turf publicly declares CLLocationCoordinate2D’s conformance to Codable and implements the protocol according to the GeoJSON standard, as an array in longitude, latitude order. But this is problematic if the application or some other library needs to support some other coordinate representation in JSON. For example, [Overpass JSON](http://overpass-api.de/output_formats.html#json) represents coordinate pairs as objects. Turf needs this Codable implementation but doesn’t need to export it publicly. It would be trivial for client code to reimplement this conformance if necessary.

https://github.com/mapbox/turf-swift/blob/1a105b11bc3acd56df71778c8769c81c033f1dcb/Sources/Turf/CoreLocation.swift#L53-L66

/cc @mapbox/navigation-ios @frederoni

PR #154, opened 2021-09-28
Compliant, codable, equatable GeoJSON

### LINKED ISSUE #150: Conform all geojson related types to Equatable
Hello,

I'm looking to store some decoded geojson values in a struct that is marked equatable. If I use Turf's types for this today, I need to manually implement the `==` operator function, which I can only do by either ignoring the geojson values entirely, or by manually specifying an `==` function for each type I want to store, and all of those types' properties' types as well. If the library could provide these conformances for us, that would be much easier.

I realize that some types' variables aren't able to easily do this, namely properties with a `[String: Any]` dictionary type. There's another library I've been using that I'd like to transition off of here (https://github.com/kiliankoe/GeoJSON) which has solved this by using an `AnyCodable` type that conforms to Equatable, so that the properties dictionary can also be Equatable. Could Turf adopt a similar approach here?

Thanks!

### LINKED ISSUE #153: FeatureCollection has nonstandard properties
[RFC 7946](https://datatracker.ietf.org/doc/html/rfc7946#section-7.1) (the GeoJSON specification) prohibits `FeatureCollection` from having a `properties` property and doesn’t call for it to have an `id `property, but this library’s GeoJSONObect protocol defines `properties` and `identifier` properties, which force `FeatureCollection` to have those properties:

https://github.com/mapbox/turf-swift/blob/6ecf55bc9553d2073e8fbb8754c4e0a27223ea4f/Sources/Turf/GeoJSON.swift#L33-L37 https://github.com/mapbox/turf-swift/blob/6ecf55bc9553d2073e8fbb8754c4e0a27223ea4f/Sources/Turf/FeatureCollection.swift#L8

We should remove these properties from both GeoJSONObject and FeatureCollection in order to conform to the GeoJSON specification. If we’re unable to make this change before v2.0.0, then we can keep both properties in Swift but encode them to JSON as foreign members by other names, such as `com.mapbox.Turf.properties` and `com.mapbox.Turf.id`.

/cc @mapbox/navigation-ios @mapbox/maps-ios

1ec5:
This PR revamps the public interface and implementation of GeoJSON support to take better advantage of the Swift compiler’s type checking capabilities. It includes a number of backwards-incompatible changes to the public API:

* Replaced the `GeoJSON` class and `GeoJSON` protocol with a unified `GeoJSONObject` enumeration. Instead of calling `GeoJSON.parse(_:)` or `GeoJSON.parse<T: GeoJSONObject>(_:from:)`, use JSONDecoder:
   ```swift
   // Before
   if let feature = try GeoJSON.parse(data)?.decodedFeature,
      case let .lineString(lineString) = feature.geometry { … }
   // Now
   if case let .feature(feature) = try JSONDecoder().decode(GeoJSONObject.self, from: data),
      case let .lineString(lineString) = feature.geometry { … }
   ```
* Removed the `FeatureCollection.identifier` and `FeatureCollection.properties` properties with no replacement. These properties had been represented in GeoJSON by foreign members, which are not yet implemented. If you had been relying on the `identifier` or `properties` foreign members of FeatureCollection objects, move the data to each individual feature in the collection.
* The `Feature.properties` property is now a `JSONObject?` (in other words, `[String: JSONValue?]?`). JSONObject is type-checked at compile time instead of runtime, but you can initialize it using a literal or full-width conversion from `Any?`. Code that builds a JSON object using literals will have to be modified to either specify a `JSONValue` case for each value or call the `JSONObject(rawValue:)` initializer.
* The `Feature.geometry` property is now optional.
* Removed the `Geometry.type` property. Use pattern matching (`case let`) instead.
* Removed the `Geometry.value` property. This type erasure is unnecessary and can potentially become a source of bugs. Use pattern matching (`case let`) instead.
* Removed the `Number` enumeration in favor of a `Double`-typed `FeatureIdentifier.number(_:)` case. JSON doesn’t distinguish between integers and double-precision floating point numbers. Any distinction in the type system or encoded JSON is purely cosmetic.
* `Feature` and `FeatureCollection` now conform to the `Equatable` protocol.
* Each geometric type, such as `Point`, now conforms to the `Codable` and `Equatable` protocols.
* `BoundingBox` now conforms to the `Hashable` protocol.

The resulting GeoJSON interface is probably much closer to other Swift implementations of GeoJSON like [GeoJSON](https://github.com/kiliankoe/GeoJSON) and [GEOSwift](https://github.com/GEOSwift/GEOSwift/). (@macdrevx gets credit for, among other things, introducing me to the `ExpressibleBy*Literal` protocols as syntactic sugar for JSON.) That said, there’s still a pragmatic reliance on Core Location types (or workalikes on Linux), such as `CLLocationCoordinate2D`/`LocationCoordinate2D` in place of a dedicated `Position` struct. Also, `null` is still represented by `Optional.none` instead of a redundant `JSONValue.null` case. Perhaps a future refactor could migrate to or vendor one of the dedicated GeoJSON libraries for more strict conformance.

In practice, most of these changes appear to be less invasive than they sound. Some downstream projects are being updated:

* mapbox-directions-swift: No change.
* mapbox/mapbox-navigation-ios#3409
* mapbox/mapbox-maps-ios#715

I’ll be the first to admit that a refactor of this size is unfortunate so close to the final release of Turf v2.0.0, but it’s the only way we can get `Equatable` conformance and better GeoJSON compliance, which is important for the wider variety of use cases associated with the map SDK compared to the navigation SDK.

Test coverage increased from 84.0% to 90.3%.

Fixes #150 and fixes #153. See #45 #96 for the previous iterations of this part of the codebase.

/cc @mapbox/navigation-ios @mapbox/maps-ios


Sources/Turf/Geometries/Point.swift
```diff
@@ -14,3 +14,27 @@ public struct Point: Equatable {
...
+        let coordinates = try container.decode(LocationCoordinate2DCodable.self, forKey: .coordinates).decodedCoordinates
+        self = .init(coordinates)
+    }
+    
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        try container.encode(Kind.Point, forKey: .kind)
+        try container.encode(coordinates.codableCoordinates, forKey: .coordinates)
+    }
+}
```
1ec5, line 40: Each of the geometric types has a very similar Codable implementation. In the interest of time, I’m leaving this redundancy in place, but it would be nice to figure out a more generic implementation at some point. The sticking point is that everything is ultimately based on `CLLocationCoordinate2D`, which we can’t conform to Codable because it would potentially collide with other libraries that use that standard type in a non-GeoJSON context: #84.

Udumft (approved):

Sources/Turf/Feature.swift
```diff
@@ -3,34 +3,50 @@ import Foundation
...
-public struct Feature: GeoJSONObject {
-    public var type: FeatureType = .feature
+/**
+ A [Feature object](https://datatracker.ietf.org/doc/html/rfc7946#section-3.2) represents a spatially bounded thing.
+ */
+public struct Feature {
     public var identifier: FeatureIdentifier?
-    public var properties: [String : Any?]?
+    public var properties: JSONObject?
     public var geometry: Geometry
```
macdrevx, line 12: Spec allows geometry to be null: https://datatracker.ietf.org/doc/html/rfc7946#section-3.2

Not a required change for this PR

Sources/Turf/Feature.swift
```diff
@@ -3,34 +3,50 @@ import Foundation
...
+    }
+}
+
+extension Feature: Equatable {
+    public static func == (lhs: Feature, rhs: Feature) -> Bool {
+        return lhs.identifier == rhs.identifier &&
+            lhs.geometry == rhs.geometry &&
+            lhs.properties == rhs.properties
+    }
+}
```
macdrevx, line 25: Can this implementation be synthesized?

Sources/Turf/FeatureIdentifier.swift
```diff
@@ -1,55 +1,37 @@
 import Foundation
 
-public enum Number: Equatable {
-    case int(Int)
-    case double(Double)
+public enum FeatureIdentifier: Equatable {
+    case string(String)
+    case number(Double)
```
macdrevx, line 5: We had a recent fix in mapbox-maps-ios related to this https://github.com/mapbox/mapbox-maps-ios/pull/676

Might be able to simplify our implementation in light of this change.

Sources/Turf/GeoJSON.swift
```diff
@@ -3,103 +3,60 @@ import Foundation
...
+ A [GeoJSON object](https://datatracker.ietf.org/doc/html/rfc7946#section-3) represents a Geometry, Feature, or collection of
+ Features.
+ */
+public enum GeoJSONObject: Equatable {
+    /**
+     A [Geometry object](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1) represents points, curves, and surfaces in coordinate space.
+     
+     - parameter geometry: The GeoJSON object as a Geometry object.
+     */
+    case geometry(_ geometry: Geometry)
```
macdrevx, line 16: Is the parameter naming just to benefit documentation?

Sources/Turf/JSON.swift
```diff
@@ -0,0 +1,183 @@
...
+        } else if let string = try? container.decode(String.self) {
+            self = .string(string)
+        } else if let object = try? container.decode(JSONObject.self) {
+            self = .object(object)
+        } else if let array = try? container.decode(JSONArray.self) {
+            self = .array(array)
+        } else {
+            throw DecodingError.typeMismatch(JSONValue.self, DecodingError.Context(codingPath: decoder.codingPath, debugDescription: "Unable to decode as a JSONValue."))
+        }
+    }
```
macdrevx, line 166: how does this decode `null`?

Sources/Turf/Feature.swift
1ec5: Oof, I guess we might as well take care of that now rather than later. It’s going to cause some more churn downstream, though any code that uses pattern matching on `geometry` will be unaffected.

Sources/Turf/FeatureIdentifier.swift
1ec5: Yep, mapbox/mapbox-maps-ios#715 simplifies the conversion you’re referring to.

Sources/Turf/GeoJSON.swift
1ec5: The parameter name also appears in the completion token when code-completing the case.

Sources/Turf/JSON.swift
```diff
@@ -0,0 +1,183 @@
...
+        } else if let number = rawValue as? NSNumber {
+            self = .number(number.doubleValue)
+        } else if let array = rawValue as? JSONArray.RawValue {
+            self = .array(.init(rawValue: array))
+        } else if let object = rawValue as? JSONObject.RawValue {
+            self = .object(.init(rawValue: object))
+        } else {
+            return nil
+        }
+    }
```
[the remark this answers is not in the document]
1ec5: `JSONArray` is typed as `[JSONValue?]`, so a `null` value is represented by `Optional.none`. The `else` statement happens to fail the initializer, resulting in `Optional.none` for any NSNull present in the input.

Sources/Turf/JSON.swift
```diff
@@ -0,0 +1,183 @@
...
+ A JSON array of `JSONValue` instances.
+ */
+public typealias JSONArray = [JSONValue?]
+
+extension JSONArray: RawRepresentable {
+    public typealias RawValue = [Any?]
+    
+    public init(rawValue values: RawValue) {
+        self = values.map(JSONValue.init(rawValue:))
+    }
```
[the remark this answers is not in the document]
1ec5: Yes, this is consistent with `RawRepresentable(rawValue:)` semantics.

Sources/Turf/JSON.swift
1ec5: It doesn’t: `null` is represented externally by `Optional.none`.

macdrevx (approved):

Sources/Turf/JSON.swift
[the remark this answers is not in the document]
1ec5: Or would it be better for this initializer to fail if any of the elements fail to initialize?

1ec5:
8aaabfffce52308ac3dcfc6ae7eb89a126a4854e updates the GeoJSON example in the readme.

PR #159, opened 2021-09-30
Add conversions between enumerations and associated values

### LINKED ISSUE #158: Geometric types should conform to common protocol that converts to Geometry
GEOSwift defines a [GeometryConvertible](https://github.com/GEOSwift/GEOSwift/blob/main/GEOSwift/Core%20Types/GeometryConvertible.swift) protocol that makes it possible to initialize a Feature without redundantly specifying both the Geometry case and the geometric type’s initializer at the same time. Turf’s GeoJSONObject and Geometry enumerations should have something similar. For example:

```swift
var geoJSON = Feature(geometry: .point(Point(coordinate)))
```

would become simply:

```swift
var geoJSON = Feature(geometry: Point(coordinate))
```

It’s similar to the initializers and literals that #154 added to JSONValue, but this protocol distributes the implementation across each type:

https://github.com/mapbox/turf-swift/blob/007896bb945432d4683e114fcf2f089edff4ceeb/Sources/Turf/JSON.swift#L49-L52

/cc @macdrevx

1ec5:
Added protocol-based initializers to `Geometry` and `GeoJSONObject` that accept each case’s associated value as an initializer argument. This makes the enumerations more consistent with the `JSONValue` and `FeatureIdentifier` enumerations and simplifies some common workflows.

JSONValue and FeatureIdentifier now conform to the `Hashable` protocol. These types are likely to be used as dictionary keys or set members, and it’s pretty easy to make them hashable.

Fixes #158.

/cc @mapbox/navigation-ios @mapbox/maps-ios

### DIFF
--- a/Sources/Turf/GeoJSON.swift
+++ b/Sources/Turf/GeoJSON.swift
@@ -74,6 +74,10 @@ public protocol GeoJSONObjectConvertible {
     var geoJSONObject: GeoJSONObject { get }
 }
 
+extension GeoJSONObject: GeoJSONObjectConvertible {
+    public var geoJSONObject: GeoJSONObject { return self }
+}
+
 extension Geometry: GeoJSONObjectConvertible {
     public var geoJSONObject: GeoJSONObject { return .geometry(self) }
 }
--- a/Sources/Turf/GeoJSON.swift
+++ b/Sources/Turf/GeoJSON.swift
@@ -28,6 +28,11 @@ public enum GeoJSONObject: Equatable {
      - parameter featureCollection: The GeoJSON object as a FeatureCollection object.
      */
     case featureCollection(_ featureCollection: FeatureCollection)
+    
+    /// Initializes a GeoJSON object representing the given GeoJSON object–convertible instance.
+    public init(_ object: GeoJSONObjectConvertible) {
+        self = object.geoJSONObject
+    }
 }
 
 extension GeoJSONObject: Codable {
@@ -59,4 +64,24 @@ extension GeoJSONObject: Codable {
             try container.encode(featureCollection)
         }
     }
+}
+
+/**
+ A type that can be represented as a `GeoJSONObject` instance.
+ */
+public protocol GeoJSONObjectConvertible {
+    /// The instance wrapped in a `GeoJSONObject` instance.
+    var geoJSONObject: GeoJSONObject { get }
+}
+
+extension Geometry: GeoJSONObjectConvertible {
+    public var geoJSONObject: GeoJSONObject { return .geometry(self) }
+}
+
+extension Feature: GeoJSONObjectConvertible {
+    public var geoJSONObject: GeoJSONObject { return .feature(self) }
+}
+
+extension FeatureCollection: GeoJSONObjectConvertible {
+    public var geoJSONObject: GeoJSONObject { return .featureCollection(self) }
 }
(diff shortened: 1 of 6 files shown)


1ec5:
[Linux builds](https://app.travis-ci.com/github/mapbox/turf-swift/builds/238810618) are failing with several errors like the following:

```
/home/travis/build/mapbox/turf-swift/Sources/Turf/GeoJSON.swift:78:48: error: reference to member 'geometry' cannot be resolved without a contextual type

    public var geoJSONObject: GeoJSONObject { .geometry(self) }

                                              ~^~~~~~~~
```

Sources/Turf/GeoJSON.swift
```diff
@@ -60,3 +65,23 @@ extension GeoJSONObject: Codable {
         }
     }
 }
+
+/**
+ A type that can be represented as a `GeoJSONObject` instance.
+ */
+public protocol GeoJSONObjectConvertible {
```
macdrevx, line 72: Below, you made `Geometry` conform to the corresponding protocol. Should we do the same here with `GeoJSONObject`?

Sources/Turf/JSON.swift
```diff
@@ -38,27 +38,22 @@ public enum JSONValue: Equatable {
      - parameter number: An integer. JSON does not distinguish numeric types of different precisions, so the integer is stored as a floating-point number.
      */
     public init<Source>(_ number: Source) where Source: BinaryInteger {
-        self = .number(Double(number))
+        self.init(Double(number))
     }
```
macdrevx, line 42: Would it be better to extend `BinaryInteger` to conform to `JSONValueConvertible`?

macdrevx: (and remove this init method?)

Sources/Turf/JSON.swift
```diff
@@ -209,3 +217,39 @@ extension JSONValue: Codable {
         }
     }
 }
+
+/**
+ A type that can be represented as a `JSONValue` instance.
+ */
+public protocol JSONValueConvertible {
```
macdrevx, line 224: Should `Optional` be `JSONValueConvertible` in situations where its generic type is too?

Sources/Turf/JSON.swift
```diff
@@ -209,3 +217,39 @@ extension JSONValue: Codable {
...
+    public var jsonValue: JSONValue { return .boolean(self) }
+}
+
+extension Array: JSONValueConvertible where Element == JSONValueConvertible? {
+    public var jsonValue: JSONValue {
+        return .array(map { $0?.jsonValue })
+    }
+}
+
+extension Dictionary: JSONValueConvertible where Key == String, Value == JSONValueConvertible? {
```
macdrevx, line 251: If we did https://github.com/mapbox/turf-swift/pull/159/files#r721695103, maybe these generic constraints could be `== JSONValueConvertible`

macdrevx (commented):
I feel really comfortable with the GeometryConvertible and GeoJSONObjectConvertible portions of this PR. It's less obvious whether the JSONValueConvertible portions are what we want. If there's a way to defer those parts, I think it'd be worth taking the extra time.

1ec5:
I split out the `JSONValueConvertible` changes as #164, since there are still open questions about the extra syntactic sugar it would introduce.

Sources/Turf/JSON.swift
1ec5: No, unfortunately that isn’t possible:

```
/path/to/turf-swift/Sources/Turf/JSON.swift:233:1: Extension of protocol 'BinaryInteger' cannot have an inheritance clause
```

Sources/Turf/GeoJSON.swift
1ec5: I finally realized what “idempotent” means. 😄

PR #166, opened 2021-10-19
Add method of LineString.trimmed(from:to:)

### LINKED ISSUE #156: Add method to slice LineString between two distances
Right now in `Turf.js`, there's a method to slice a `LineString` based on the start and end distance, as [TurfMisc.lineSliceAlong(line: startDist: stopDist: options)](https://github.com/Turfjs/turf/tree/master/packages/turf-line-slice-along/).

And right now in `Turf.swift`, we have the `LineString.trimmed(from:distance)` to slice a `LineString` based on the start coordinate and the distance.

So is there any way to slice a `LineString` based on the start and end distance? Right now I'm implementing the following method to do so, but not sure whether it would get the same result as `TurfMisc.lineSliceAlong(line: lineString, startDist: startDistance, stopDist: stopDistance)`

```
if let midPoint = lineString.coordinateFromStart(distance: startDistance) {
    let slicedLine = lineString.trimmed(from: midPoint, distance: stopDistance - startDistance)
}
```

ShanMa1991:
To fix #156 , adding the method of `LineString.trimmed(from:to:) ` , to return a sliced `LineString` based on the starting distance and stop distance.

### DIFF
--- a/Sources/Turf/Geometries/LineString.swift
+++ b/Sources/Turf/Geometries/LineString.swift
@@ -78,9 +78,63 @@ extension LineString {
     }
     
     /**
-     Returns the portion of the line string that begins at the given coordinate and extends the given distance along the line string.
+     Returns the portion of the line string that begins at the given start distance and extends the given stop distance along the line string.
      
-     This method is roughly equivalent to the [turf-line-slice-along](https://turfjs.org/docs/#lineSliceAlong) package of Turf.js ([source code](https://github.com/Turfjs/turf/tree/master/packages/turf-line-slice-along/)), except that it accepts a starting position instead of a starting distance along the line string.
+     This method is roughly equivalent to the [turf-line-slice-along](https://turfjs.org/docs/#lineSliceAlong) package of Turf.js ([source code](https://github.com/Turfjs/turf/tree/master/packages/turf-line-slice-along/)).
+     */
+    public func lineSliceAlong(startDistance: LocationDistance, stopDistance: LocationDistance) -> LineString? {
+        let coordinates = self.coordinates
+        var traveled: LocationDistance = 0
+        var slice = [LocationCoordinate2D]()
+        
+        for i in 0..<coordinates.endIndex {
+            if startDistance >= traveled && i == coordinates.endIndex - 1 {
+                break
+            } else if traveled > startDistance && slice.isEmpty {
+                let overshoot = startDistance - traveled
+                if overshoot == 0.0 {
+                    slice.append(coordinates[i])
+                    return LineString(slice)
+                }
+                let direction = coordinates[i].direction(to: coordinates[i - 1]) - 180
+                let interpolated = coordinates[i].coordinate(at: overshoot, facing: direction)
+                slice.append(interpolated)
+            }
+            
+            if traveled >= stopDistance {
+                let overshoot = stopDistance - traveled
+                if overshoot == 0.0 {
+                    slice.append(coordinates[i])
+                    return LineString(slice)
+                }
+                let direction = coordinates[i].direction(to: coordinates[i - 1]) - 180
+                let interpolated = coordinates[i].coordinate(at: overshoot, facing: direction)
+                slice.append(interpolated)
+                return LineString(slice)
+            }
+            
+            if traveled >= startDistance {
+                slice.append(coordinates[i])
+            }
+            
+            if i == coordinates.count - 1 {
+                return LineString(slice)
+            }
+            
+            traveled += distance(from: coordinates[i], to: coordinates[i + 1]) ?? 0.0
+        }
+        
+        if traveled < startDistance { return nil }
+        
+        if let last = coordinates.last {
+            return LineString([last, last])
+        }
+        
+        return nil
+    }
+    
+    /**
+     Returns the portion of the line string that begins at the given coordinate and extends the given distance along the line string.
      */
     public func trimmed(from coordinate: LocationCoordinate2D, distance: LocationDistance) -> LineString? {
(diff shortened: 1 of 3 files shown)


Sources/Turf/Geometries/LineString.swift
```diff
@@ -78,9 +78,63 @@ extension LineString {
...
     
     /**
-     Returns the portion of the line string that begins at the given coordinate and extends the given distance along the line string.
+     Returns the portion of the line string that begins at the given start distance and extends the given stop distance along the line string.
      
-     This method is roughly equivalent to the [turf-line-slice-along](https://turfjs.org/docs/#lineSliceAlong) package of Turf.js ([source code](https://github.com/Turfjs/turf/tree/master/packages/turf-line-slice-along/)), except that it accepts a starting position instead of a starting distance along the line string.
+     This method is roughly equivalent to the [turf-line-slice-along](https://turfjs.org/docs/#lineSliceAlong) package of Turf.js ([source code](https://github.com/Turfjs/turf/tree/master/packages/turf-line-slice-along/)).
+     */
+    public func lineSliceAlong(startDistance: LocationDistance, stopDistance: LocationDistance) -> LineString? {
+        let coordinates = self.coordinates
```
1ec5, line 86: When porting a method from Turf.js, add a comment to the beginning that has the permalink to the Turf.js source code to make it easier to compare the ported code to the original code.

Sources/Turf/Geometries/LineString.swift
```diff
@@ -78,9 +78,63 @@ extension LineString {
...
+        var traveled: LocationDistance = 0
+        var slice = [LocationCoordinate2D]()
+        
+        for i in 0..<coordinates.endIndex {
```
1ec5, line 90: This is a much more literal port of the JavaScript method than the existing method below. Do we know if the version below was correct, apart from the difference in inputs? We should find a way for one method to be based on the other.

Sources/Turf/Geometries/LineString.swift
```diff
@@ -78,9 +78,63 @@ extension LineString {
     }
     
     /**
-     Returns the portion of the line string that begins at the given coordinate and extends the given distance along the line string.
+     Returns the portion of the line string that begins at the given start distance and extends the given stop distance along the line string.
      
-     This method is roughly equivalent to the [turf-line-slice-along](https://turfjs.org/docs/#lineSliceAlong) package of Turf.js ([source code](https://github.com/Turfjs/turf/tree/master/packages/turf-line-slice-along/)), except that it accepts a starting position instead of a starting distance along the line string.
+     This method is roughly equivalent to the [turf-line-slice-along](https://turfjs.org/docs/#lineSliceAlong) package of Turf.js ([source code](https://github.com/Turfjs/turf/tree/master/packages/turf-line-slice-along/)).
+     */
+    public func lineSliceAlong(startDistance: LocationDistance, stopDistance: LocationDistance) -> LineString? {
```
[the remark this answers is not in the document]
ShanMa1991: Thanks for the suggestion, I'll update the name and add the permalink to the Turf.js source code.

Sources/Turf/Geometries/LineString.swift
ShanMa1991: The method below is different from the `trimmed(from:to:)` because it choose a `closestCoordinate` as a start point instead of a start distance. If we convert the `closestCoordinate` to the start distance and then use the `trimmed(from:to:)`, it would lead to more calculation than the original one. 

And I tried to use the below method to replace the `trimmed(from:to:)` , but it could not pass some tests. And the `trimmed(from:to:)` has much less calculation than the twice use of `trimmed(from:distance:)`. 

So I think they two are better kept separate for different use cases.

Sources/Turf/Geometries/LineString.swift
```diff
@@ -78,9 +78,65 @@ extension LineString {
...
+        
+        for i in 0..<coordinates.endIndex {
+            if startDistance >= traveled && i == coordinates.endIndex - 1 {
+                break
+            } else if traveled > startDistance && slice.isEmpty {
+                let overshoot = startDistance - traveled
+                if overshoot == 0.0 {
+                    slice.append(coordinates[i])
+                    return LineString(slice)
+                }
```
1ec5, line 100: This looks very similar to `indexedCoordinateFromStart(distance:)`, which was ported from turf-along. Is there an opportunity to reuse that method?

https://github.com/mapbox/turf-swift/blob/33da423c56216c4dc41c68beb130019081085113/Sources/Turf/Geometries/LineString.swift#L162-L179

ShanMa1991: Theoretically, we could using the following the avoid the code duplication:
```
guard startDistance >= 0.0 && stopDistance >= startDistance else { return nil }
guard let startIndexedCoordinate = indexedCoordinateFromStart(distance: startDistance) else { return nil }
return trimmed(from: startIndexedCoordinate.coordinate, distance: stopDistance - startDistance)
```
But in fact, we shouldn't do that. The above failed several unit tests with the following reasons:
1. When the startDistance is larger than the lineString distance, it should return nil but in fact it would return the last coordinate.
2. The `trimmed(from:distance:)` doesn't accept the `startIndexedCoordinate.coordinate` directly, but in fact it would call `closestCoordinate(to:)` again to find the coordinate. It would lead to the wrong start coordinate and possible wrong end coordinate, when the stop distance is smaller than the whole distance of LineString.

Except the test failure, it would also increase the calculation, because the `closestCoordinate(to:)` would loop through the lineString's coordinates for one more time.

1ec5 (approved):

PR #167, opened 2021-10-29
Linestring.intersection method

### LINKED ISSUE #141: Add LineString intersection method
`intersection(_:_:)` takes two instances of `LineSegment`, which is just a type alias for a tuple of two coordinates. This is an unusual type used nowhere else in the library. I think it’s intended to be more efficient than `LineString` for intersection purposes, but `LineString` would be more discoverable. We should add a `LineString.intersection(with:)` method that takes another `LineString`. This method would be more discoverable and intuitive than the existing `intersection(_:_:)` method, and we might eventually have an opportunity to make it more generic as part of a protocol once #51 is implemented.

/cc @mapbox/navigation-ios @ZiZasaurus

Udumft:
Resolves #141 

PR adds zLinestring.intersection(with:)` method.

### DIFF
--- a/Sources/Turf/Geometries/LineString.swift
+++ b/Sources/Turf/Geometries/LineString.swift
@@ -308,7 +308,7 @@ extension LineString {
             let direction = segment.0.direction(to: segment.1)
             let perpendicularPoint1 = coordinate.coordinate(at: maxDistance, facing: direction + 90)
             let perpendicularPoint2 = coordinate.coordinate(at: maxDistance, facing: direction - 90)
-            let intersectionPoint = intersection((perpendicularPoint1, perpendicularPoint2), segment)
+            let intersectionPoint = Turf.intersection((perpendicularPoint1, perpendicularPoint2), segment)
             let intersectionDistance: LocationDistance? = intersectionPoint != nil ? coordinate.distance(to: intersectionPoint!) : nil
             
             if distances.0 < closestDistance ?? .greatestFiniteMagnitude {
@@ -363,5 +363,26 @@ extension LineString {
     public mutating func simplify(tolerance: Double = 1.0, highestQuality: Bool = false) {
         // Ported from https://github.com/Turfjs/turf/blob/4e8342acb1dbd099f5e91c8ee27f05fb2647ee1b/packages/turf-simplify/lib/simplify.js
         coordinates = Simplifier.simplify(coordinates, tolerance: tolerance, highestQuality: highestQuality)
+    }
+    
+    /**
+     Returns all intersections with another `LineString`.
+     
+     This function is roughly equivalent to the [turf-line-intersect](https://turfjs.org/docs/#lineIntersect) package of Turf.js ([source code](https://github.com/Turfjs/turf/tree/master/packages/turf-line-intersect/)).
+     */
+    public func intersection(with line: LineString) -> [LocationCoordinate2D]? {
+        var intersections: [String : LocationCoordinate2D] = [:]
+        zip(coordinates.dropLast(), coordinates.dropFirst()).forEach { segment1 in
+            zip(line.coordinates.dropLast(), line.coordinates.dropFirst()).forEach { segment2 in
+                if let intersection = Turf.intersection(LineSegment(segment1.0, segment1.1),
+                                                        LineSegment(segment2.0, segment2.1)) {
+                    let key = "\(intersection.latitude),\(intersection.longitude)"
+                    if intersections[key] == nil {
+                        intersections[key] = intersection
+                    }
+                }
+            }
+        }
+        return intersections.isEmpty ? ([LocationCoordinate2D]?).none : Array(intersections.values)
     }
 }
--- a/Sources/Turf/Geometries/LineString.swift
+++ b/Sources/Turf/Geometries/LineString.swift
@@ -382,19 +382,30 @@ extension LineString {
      - seealso: `Turf.intersection(_:, _:)`
      */
     public func intersections(with line: LineString) -> [LocationCoordinate2D] {
-        var intersections: [String : LocationCoordinate2D] = [:]
-        
+        var intersections = Set<HashableCoordinate>()
         for segment1 in segments {
             for segment2 in line.segments {
                 if let intersection = Turf.intersection(LineSegment(segment1.0, segment1.1),
                                                         LineSegment(segment2.0, segment2.1)) {
-                    let key = "\(intersection.latitude),\(intersection.longitude)"
-                    if intersections[key] == nil {
-                        intersections[key] = intersection
-                    }
+                    intersections.insert(.init(intersection))
                 }
             }
         }
-        return Array(intersections.values)
+        return intersections.map(\.locationCoordinate)
+    }
+    
+    private struct HashableCoordinate: Hashable {
+        let latitude: Double
+        let longitude: Double
+        
+        var locationCoordinate: LocationCoordinate2D {
+            return LocationCoordinate2D(latitude: latitude,
+                                        longitude: longitude)
+        }
+        
+        init(_ coordinate: LocationCoordinate2D) {
+            self.latitude = coordinate.latitude
+            self.longitude = coordinate.longitude
+        }
     }
 }
--- a/Sources/Turf/Geometries/LineString.swift
+++ b/Sources/Turf/Geometries/LineString.swift
@@ -391,7 +391,7 @@ extension LineString {
                 }
             }
         }
-        return intersections.map(\.locationCoordinate)
+        return intersections.map { $0.locationCoordinate }
     }
     
     private struct HashableCoordinate: Hashable {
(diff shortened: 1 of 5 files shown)


Sources/Turf/Geometries/LineString.swift
```diff
@@ -308,4 +308,25 @@ extension LineString {
         // Ported from https://github.com/Turfjs/turf/blob/4e8342acb1dbd099f5e91c8ee27f05fb2647ee1b/packages/turf-simplify/lib/simplify.js
         coordinates = Simplifier.simplify(coordinates, tolerance: tolerance, highestQuality: highestQuality)
     }
+    
+    /**
+     Returns all intersections with another `LineString`.
+     
+     This function is roughly equivalent to the [turf-line-intersect](https://turfjs.org/docs/#lineIntersect) package of Turf.js ([source code](https://github.com/Turfjs/turf/tree/master/packages/turf-line-intersect/)).
```
1ec5, line 315: Mention `intersection(_:_:)` as an alternative for working with individual line segments.

Update the correspondence table in the readme to link to this method:

https://github.com/mapbox/turf-swift/blob/1806393ee69e219ddc30ebf16cab3afc0436fe41/README.md#L90

Sources/Turf/Geometries/LineString.swift
```diff
@@ -308,4 +308,25 @@ extension LineString {
...
+     */
+    public func intersection(with line: LineString) -> [LocationCoordinate2D]? {
+        var intersections: [String : LocationCoordinate2D] = [:]
+        zip(coordinates.dropLast(), coordinates.dropFirst()).forEach { segment1 in
+            zip(line.coordinates.dropLast(), line.coordinates.dropFirst()).forEach { segment2 in
```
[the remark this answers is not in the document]
1ec5: > [Original implementation](https://github.com/Turfjs/turf/blob/master/packages/turf-line-intersect/index.ts#L68-L70) has tree search for corresponding segments

Turf.js’s use of an optimized R-tree from [RBush](https://github.com/mourner/rbush/) is key to the method’s high performance. There is a similar [GKRTree](https://developer.apple.com/documentation/gameplaykit/gkrtree) in GameplayKit, and apparently some support in Core Data as well, but I’m not sure we’d want to maintain a separate codepath for non-Apple platforms. I also haven’t found any good RTree implementations in pure Swift yet; [this package](https://github.com/emma-k-alexandra/RTree/) stores the tree inefficiently on disk. Let’s ticket out a port of RBush separately as a performance optimization.

Sources/Turf/Geometries/LineString.swift
```diff
@@ -308,4 +308,25 @@ extension LineString {
...
+                if let intersection = Turf.intersection(LineSegment(segment1.0, segment1.1),
+                                                        LineSegment(segment2.0, segment2.1)) {
+                    let key = "\(intersection.latitude),\(intersection.longitude)"
```
1ec5, line 323: A string representation is kind of weird, but it’s probably the most straightforward approach until apple/swift#37398 lands, since we wouldn’t be able to have `LocationCoordinate2D` privately conform to `Hashable`.

Sources/Turf/Geometries/LineString.swift
```diff
@@ -308,4 +308,25 @@ extension LineString {
         // Ported from https://github.com/Turfjs/turf/blob/4e8342acb1dbd099f5e91c8ee27f05fb2647ee1b/packages/turf-simplify/lib/simplify.js
         coordinates = Simplifier.simplify(coordinates, tolerance: tolerance, highestQuality: highestQuality)
     }
+    
+    /**
+     Returns all intersections with another `LineString`.
+     
+     This function is roughly equivalent to the [turf-line-intersect](https://turfjs.org/docs/#lineIntersect) package of Turf.js ([source code](https://github.com/Turfjs/turf/tree/master/packages/turf-line-intersect/)).
+     */
+    public func intersection(with line: LineString) -> [LocationCoordinate2D]? {
```
1ec5, line 317: This method can return a whole array of intersections, so rename it to `intersections(with:)`.

Sources/Turf/Geometries/LineString.swift
```diff
@@ -308,4 +308,25 @@ extension LineString {
...
+        var intersections: [String : LocationCoordinate2D] = [:]
+        zip(coordinates.dropLast(), coordinates.dropFirst()).forEach { segment1 in
```
1ec5, line 319: Factor out this `zip(_:_:)` call as `LineString.segments` for readability:

```swift
for segment1 in segments {
    for segment2 in line.segments {
```

Sources/Turf/Geometries/LineString.swift
```diff
@@ -308,4 +308,25 @@ extension LineString {
...
+                if let intersection = Turf.intersection(LineSegment(segment1.0, segment1.1),
+                                                        LineSegment(segment2.0, segment2.1)) {
+                    let key = "\(intersection.latitude),\(intersection.longitude)"
+                    if intersections[key] == nil {
+                        intersections[key] = intersection
+                    }
+                }
+            }
+        }
+        return intersections.isEmpty ? ([LocationCoordinate2D]?).none : Array(intersections.values)
```
1ec5, line 330: If there are no intersections, return an empty array instead of `nil`, to simplify the set of possible return values.

Sources/Turf/Geometries/LineString.swift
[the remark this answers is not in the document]
Udumft: Added a [ticket](https://github.com/mapbox/turf-swift/issues/171) to track porting.

1ec5 (approved):

Sources/Turf/Geometries/LineString.swift
bamx23: I'm a bit concerned about floating-point accuracy here. And about time complexity constant.
What if we implement a small struct like
```swift
private struct HashableCoordinate: Hashable {
    var latitude: Double
    var longitude: Double
}
```
And use `Set<HashableCoordinate>` that is then converted to an array of coordinates.
Or is it too much? May it be useful somewhere else in Turf?

bamx23 (approved):

Sources/Turf/Geometries/LineString.swift
1ec5: Yes, that would work. Pedantically, Turf should represent GeoJSON Point objects as its own hashable `Point` type instead of reusing `CLLocationCoordinate2D`. We stuck to the latter type as a practical matter, to avoid excessive conversions to and from a very common type. This intersection method’s performance would be dominated by the non-RTree-based intersection search, but you have a point that the struct representation would give us more certainty.

Sources/Turf/Turf.swift
```diff
@@ -35,9 +37,9 @@ public func intersection(_ line1: LineSegment, _ line2: LineSegment) -> Location
                                             longitude: line1.0.longitude + a * (line1.1.longitude - line1.0.longitude))
     
     /// True if line 1 is finite and line 2 is infinite.
-    let intersectsWithLine1 = a > 0 && a < 1
+    let intersectsWithLine1 = a >= 0 && a <= 1
```
Udumft, line 40: I am not sure if this was intended or not, but strict comparison here will not detect an intersection by line's ending points (for example when 2 LineStrings share the same coordinate). [turf-line-intersect](https://github.com/Turfjs/turf/blob/master/packages/turf-line-intersect/index.ts#L124) has such detection, but `Turfjs`'s `point-on-line` implementation (which seems to be used as base reference for Swift port) for some reason [does not](https://github.com/Turfjs/turf/blob/142e137ce0c758e2825a260ab32b24db0aa19439/packages/turf-point-on-line/index.js#L138). It seems like a bug to me, as when I seek for intersections between lines I usually want to include end points, especially if it can happen even with LineString's intermediate points, so I decided to fix that aswell.

Sources/Turf/Geometries/LineString.swift
Udumft: Implemented a Hashable wrapper

Tests/TurfTests/LineStringTests.swift
```diff
@@ -350,4 +350,21 @@ class LineStringTests: XCTestCase {
...
+    func testIntersections() {
+        let lineString = LineString([.init(latitude: 2, longitude: 1),
+                                     .init(latitude: 2, longitude: 5),
+                                     .init(latitude: 2, longitude: 9)])
+        
+        let intersectingLineString = LineString([.init(latitude: 4, longitude: 1),
+                                                 .init(latitude: 0, longitude: 5),
+                                                 .init(latitude: 2, longitude: 9)])
+        
+        let intersections = lineString.intersections(with: intersectingLineString)
```
bamx23, line 363: Should we test edge cases? Like "empty LineString" and "no intersections"?

bamx23:
LGTM

Issue #173, opened 2021-12-07
Int property of value = 1 converts to bool unexpectedly


metaine:
```
 if let bool = rawValue as? Bool {
            self = .boolean(bool)
   } 
```

1ec5:
https://github.com/mapbox/turf-swift/blob/436682278bd65a1dbfbad53902af6f4cc0af1a33/Sources/Turf/JSON.swift#L70-L75

1ec5:
@metaine, can you provide example code that results in the unexpected conversion? I added a unit test in 45213339c62c6afc88c5099c8adf43e03509867d demonstrating that `JSONValue(rawValue:)` keeps Booleans as Booleans and integers as numbers, but maybe you’re running into an issue somewhere else?

PR #181, opened 2022-02-25
Only unbox NSNumbers 0, 1 to Bool for objCType 'c'

macdrevx:
Previously, initializing `JSONValue` with `NSNumber` with value 0 or 1 would produce `JSONValue.boolean`. This change updates the behavior to only interpret `NSNumber` as boolean if its `objCType` is `'c'` and its value is 0 or 1.

### DIFF
--- a/Sources/Turf/JSON.swift
+++ b/Sources/Turf/JSON.swift
@@ -67,12 +67,42 @@ extension JSONValue: RawRepresentable {
     
     public init?(rawValue: Any) {
         // Like `JSONSerialization.jsonObject(with:options:)` with `JSONSerialization.ReadingOptions.fragmentsAllowed` specified.
-        if let bool = rawValue as? Bool {
-            self = .boolean(bool)
-        } else if let string = rawValue as? String {
+        if let string = rawValue as? String {
             self = .string(string)
         } else if let number = rawValue as? NSNumber {
-            self = .number(number.doubleValue)
+            /// When a Swift Bool or Objective-C BOOL is boxed with NSNumber, the value of the
+            /// resulting NSNumber's objCType property is 'c' (Int8 (aka CChar) in Swift, char in
+            /// Objective-C) and the value is 0 for false/NO and 1 for true/YES.
+            ///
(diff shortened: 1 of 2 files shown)


Tests/TurfTests/JSONTests.swift
```diff
@@ -8,6 +8,10 @@ class JSONTests: XCTestCase {
         XCTAssertEqual(JSONValue(rawValue: 3.1415 as NSNumber), .number(3.1415))
         XCTAssertEqual(JSONValue(rawValue: false as NSNumber), .boolean(false))
         XCTAssertEqual(JSONValue(rawValue: true as NSNumber), .boolean(true))
+        XCTAssertEqual(JSONValue(rawValue: false), .boolean(false))
+        XCTAssertEqual(JSONValue(rawValue: true), .boolean(true))
```
macdrevx, line 12: Not sure whether supporting this was intentional @1ec5, but the old implementation did allow it, so I kept the behavior and added tests to verify it.

Sources/Turf/JSON.swift
```diff
@@ -67,12 +67,42 @@ extension JSONValue: RawRepresentable {
...
+            /// initializer). Moreover, NSNumber doesn't guarantee to preserve the type suggested
+            /// by the initializer that's used to create it.
+            ///
+            /// This means that when these values are encountered, it is ambiguous whether to
+            /// decode to JSONValue.number or JSONValue.boolean.
+            ///
+            /// In practice, choosing .boolean yields the desired result more often since it is more
+            /// common to work with Bool than it is Int8.
+            switch String(cString: number.objCType) {
+            case "c": // char
```
1ec5, line 88: Thanks, this is consistent with https://github.com/mapbox/mapbox-gl-native/pull/6049#discussion_r77057040. so I’m confident in this fix.

1ec5 (approved):
Thanks, your solution demonstrates what I had been missing in https://github.com/mapbox/turf-swift/issues/173#issuecomment-1031830693.

PR #187, opened 2022-04-11
Optional foreign members coding

### LINKED ISSUE #186: Optionally avoid decoding or encoding foreign members
ForeignMemberContainer’s Codable implementation can be very slow for complex GeoJSON because it needs to iterate over `allKeys` and convert each individual coding key. The Codable implementations of ForeignMemberContainer-containing types should short-circuit foreign member encoding and decoding when the encoder or decoder’s `userInfo` contains a flag suppressing this step.

/ref https://github.com/mapbox/mapbox-directions-swift/pull/669#discussion_r842877486
/cc @mapbox/navigation-ios @macdrevx

Udumft:
Resolved #186 
This PR adds `userInfo` flag which enables foreign members coding. This changes existing behavior as previously this feature was enabled by default.

### DIFF
--- a/Sources/Turf/GeoJSON.swift
+++ b/Sources/Turf/GeoJSON.swift
@@ -103,6 +103,9 @@ extension ForeignMemberContainer {
      Decodes any foreign members using the given decoder.
      */
     mutating func decodeForeignMembers<WellKnownCodingKeys>(notKeyedBy _: WellKnownCodingKeys.Type, with decoder: Decoder) throws where WellKnownCodingKeys: CodingKey {
+        guard let allowCoding = decoder.userInfo[.geometryForeignMembersCodingKey] as? Bool,
+              allowCoding else { return }
+        
         let foreignMemberContainer = try decoder.container(keyedBy: AnyCodingKey.self)
         for key in foreignMemberContainer.allKeys {
             if WellKnownCodingKeys(stringValue: key.stringValue) == nil {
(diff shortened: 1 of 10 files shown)


Sources/Turf/GeoJSON.swift
```diff
@@ -98,11 +98,26 @@ public protocol ForeignMemberContainer {
...
 }
 
+/**
+ Key to pass to populate a `userInfo` dictionary, which is passed to the `JSONDecoder` or `JSONEncoder` to enable processing foreign members.
+*/
+public extension CodingUserInfoKey {
+    /**
+     Indicates if coding of foreign members is enabled.
+     
+     Boolean flag to enable coding. Default (or missing) value is to ignore foreign members.
```
1ec5, line 108: Especially because of this default, we should make the key as discoverable as possible. Let’s mention it in the documentation comments for ForeignMemberContainer and the types that are most likely to contain foreign members: Feature, FeatureCollection, and GeoJSONObject.

Sources/Turf/GeoJSON.swift
```diff
@@ -98,11 +98,26 @@ public protocol ForeignMemberContainer {
...
+     */
+    static let allowForeignMembersCoding = CodingUserInfoKey(rawValue: "com.mapbox.turf.coding.allowForeignMembersCodingKey")!
```
1ec5, line 110: `includesForeignMembers` would be more succinct.

1ec5 (approved):

Sources/Turf/GeoJSON.swift
Udumft: Updated

Sources/Turf/GeoJSON.swift
Udumft: Added notes for `GeoJSONObject` and `ForeignMemberContainer.foreignMembers`. `Feature` and `FeatureCollection` have doc automatically imported from `ForeignMemberContainer`.

PR #191, opened 2022-05-24
v2.5.0

ShanMa1991:
This Pr is to update Turf for Swift to `v2.5.0`


1ec5:
We don’t need a v2.5.0 release, because nothing has landed since v2.4.0. This does mean Turf versioning will once again depart from MapboxNavigation versioning, but there’s plenty of precedent for that. After all, MapboxSpeech is still on v2.0.0.

PR #192, opened 2022-06-16
Improve performance of trimmed(from:,to:)

jonkan:
So, I found the performance of the `LineString.trimmed(from:to:)` to be unusably slow and that it's caused by this distance calculation.

Calculating the distance between coordinates `i` and `i+1` can be done with a simple/cheap `distance(to:)`, but instead the `LineString.distance(from:to:)` is used, which kills performance because it allows `from`/`to` to not be on the line and has to calculate which line coordinate is closest using `closestCoordinate(to:)`, a O(n) operation, which leaves `trimmed(from:to:)` at a terrible O(n^2).

Please let me know if I'm missing anything.

### DIFF
--- a/Sources/Turf/Geometries/LineString.swift
+++ b/Sources/Turf/Geometries/LineString.swift
@@ -134,7 +134,7 @@ extension LineString {
                 return LineString(slice)
             }
             
-            traveled += distance(from: coordinates[i], to: coordinates[i + 1]) ?? 0.0
+            traveled += coordinates[i].distance(to: coordinates[i + 1])
         }
         
         if traveled < startDistance { return nil }


azarovalex:
@jonkan Thanks for contributing!

Nice catch, looks like we've used the wrong `distance` method the whole time. Original turf-js repo uses [this](https://github.com/Turfjs/turf/blob/b9b338f4f06898589dc200a09712ea6709df3114/packages/turf-distance/index.ts#L30) method for distance and that's exactly the method you use in this PR.

@1ec5 please take a final look at the PR and let's merge this.

1ec5 (approved):
Great catch – this is a tricky part of porting anything from Turf.js to turf-swift, since turf-swift uses object-oriented methods instead of global functions. Thank you for your contribution!

PR #197, opened 2022-08-26
Combine all framework target into one

OdNairy:
Xcode has an option to build different platforms based on a single target. 
This kind of setup would simplify support for cross-platform apps/SDKs.

From the technical point of view, I've changed `SUPPORTED_PLATFORMS` build setting into cross-platform value:  `iphonesimulator iphoneos macosx appletvos appletvsimulator watchsimulator watchos`. SDKROOT was dropped in favor of missing value so Xcode can detect the corresponding SDK based on destination.

I also updated the Bitrise workflow to target the new scheme. I suppose it may be a build issue on different branches until this PR is merged as we store config on bitrise website instead of per-branch.


1ec5 (approved):
Somehow I missed that Carthage supports universal frameworks: #622. We’ve been setting up all the other Swift service projects the same way over the years without realizing we could get away with a universal framework.

PR #199, opened 2022-10-21
Turf `v2.6.0-beta.1`.

MaximAlien:
Turf `v2.6.0-beta.1`.


1ec5 (approved):
Remember to mention in the release notes that the minimum deployment target has increased for iOS 11 and equivalent. This won’t be a problem for Mapbox Maps SDK v10, but it could otherwise catch some users of Mapbox or MapLibre Maps SDK v6 by surprise.

PR #201, opened 2022-11-03
Remove unused schemes.

MaximAlien:
Remove unused schemes.


ShanMa1991 (approved):
I think we could also update the readme doc for the change?

1ec5 (approved):

1ec5:
> I think we could also update the readme doc for the change?

The readme doesn’t mention schemes at all. We should mention the change in the release notes:

* Fixed a build failure when installing this library using Carthage.

PR #204, opened 2022-12-13
Geometry WKT coding

Udumft:
This PR adds limited parsing mechanism for Well Known Text format to decode and encode geometry objects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,15 +144,15 @@ turf-swift contains a minimal WKT encoding/decoding support for geometries imple
 let wktString = "POINT(123.53 -12.12)"
 
 // Decoding is done using an init method
-let point = Point(fromWKT: wktString)
-let geometry = Geometry(fromWKT: wktString)
+let point = try? Point(wkt: wktString)
+let geometry = try? Geometry(wkt: wktString)
 
-print(point.coordinates)
+print(point?.coordinates)
 
 // ...
 
 // Geometries then can be serialized using a property getter
-let serializedWKTString = point?.WKTString
+let serializedWKTString = point?.wkt
 print(serializedWKTString)
 
 ```
--- a/README.md
+++ b/README.md
@@ -135,3 +135,24 @@ print(json)
 */
 
 ```
+
+## Well Known Text (WKT)
+
+turf-swift contains a minimal WKT encoding/decoding support for geometries implementing `WKTCodable` protocol.
+
+```swift
+let wktString = "POINT(123.53 -12.12)"
+
+// Decoding is done using an init method
+let point = Point(fromWKT: wktString)
+let geometry = Geometry(fromWKT: wktString)
+
+print(point.coordinates)
+
+// ...
+
+// Geometries then can be serialized using a property getter
+let serializedWKTString = point?.WKTString
+print(serializedWKTString)
+
+```
--- a/Sources/Turf/Consumer/Consumer+WKT.swift
+++ b/Sources/Turf/Consumer/Consumer+WKT.swift
@@ -1,5 +1,7 @@
 import Foundation
+#if !os(Linux)
 import CoreLocation
+#endif
 
 
 struct WKTParser {
@@ -221,57 +223,57 @@ struct WKTParser {
             guard let coords = values as? [Double] else {
                 throw WKTError.coordinatesParsingFailed(values)
             }
-            return CLLocationCoordinate2D(latitude: coords[1],
+            return LocationCoordinate2D(latitude: coords[1],
                                           longitude: coords[0])
         case .coordinatesArray:
-            guard let coords = values as? [CLLocationCoordinate2D] else {
+            guard let coords = values as? [LocationCoordinate2D] else {
                 throw WKTError.coordinatesParsingFailed(values)
             }
             return coords
         case .coordinatesArray2D:
-            guard let coords = values as? [[CLLocationCoordinate2D]] else {
+            guard let coords = values as? [[LocationCoordinate2D]] else {
                 throw WKTError.coordinatesParsingFailed(values)
             }
             return coords
         case .coordinatesArray3D:
-            guard let coords = values as? [[[CLLocationCoordinate2D]]] else {
+            guard let coords = values as? [[[LocationCoordinate2D]]] else {
                 throw WKTError.coordinatesParsingFailed(values)
             }
             return coords
         case .point:
             guard values.count > 1 else { return nil }
-            guard let coords = values[1] as? CLLocationCoordinate2D else {
+            guard let coords = values[1] as? LocationCoordinate2D else {
                 throw WKTError.coordinatesParsingFailed(values)
             }
             return Point(coords)
         case .multiPoint:
             guard values.count > 1 else { return nil }
-            let coords = values[1..<values.endIndex].compactMap { $0 as? CLLocationCoordinate2D }
+            let coords = values[1..<values.endIndex].compactMap { $0 as? LocationCoordinate2D }
             guard coords.count == values.count - 1 else {
                 throw WKTError.coordinatesParsingFailed(values)
             }
             return MultiPoint(coords)
         case .lineString:
             guard values.count > 1 else { return nil }
-            guard let coords = values[1] as? [CLLocationCoordinate2D] else {
+            guard let coords = values[1] as? [LocationCoordinate2D] else {
                 throw WKTError.coordinatesParsingFailed(values)
             }
             return LineString(coords)
         case .multiLineString:
             guard values.count > 1 else { return nil }
-            guard let coords = values[1] as? [[CLLocationCoordinate2D]] else {
+            guard let coords = values[1] as? [[LocationCoordinate2D]] else {
                 throw WKTError.coordinatesParsingFailed(values)
             }
             return MultiLineString(coords)
         case .polygon:
             guard values.count > 1 else { return nil }
-            guard let coords = values[1] as? [[CLLocationCoordinate2D]] else {
+            guard let coords = values[1] as? [[LocationCoordinate2D]] else {
                 throw WKTError.coordinatesParsingFailed(values)
             }
             return Polygon(coords)
         case .multiPolygon:
             guard values.count > 1 else { return nil }
-            guard let coords = values[1] as? [[[CLLocationCoordinate2D]]] else {
+            guard let coords = values[1] as? [[[LocationCoordinate2D]]] else {
                 throw WKTError.coordinatesParsingFailed(values)
             }
             return MultiPolygon(coords)
--- a/Sources/Turf/Consumer/Consumer+WKT.swift
+++ b/Sources/Turf/Consumer/Consumer+WKT.swift
@@ -9,7 +9,10 @@ struct WKTParser {
     
     mutating func parse(_ input: String) throws -> Any {
         let match = try wktConsumer.match(input.uppercased())
-        return try match.transform(wktTransform)!
+        guard let output = try match.transform(wktTransform) else {
+            throw WKTError.emptyOutput
+        }
+        return output
     }
     
     static func parse(_ input: String) throws -> Any {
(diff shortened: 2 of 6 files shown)


Sources/Turf/Consumer/Consumer+WKT.swift
```diff
@@ -0,0 +1,322 @@
...
+            return nil
+        case .number:
+            guard let numbers = (values as? [String])?.compactMap({ Double($0) }),
+                  numbers.count == values.count,
+                  let number = numbers.first else {
+                throw WKTError.numberParsingFailed(values)
+            }
+            return number
+        case .coordinate:
+            guard let coords = values as? [Double] else {
```
kried, line 227: Seems like we also need to check the number of elements in the values.

Sources/Turf/Consumer/Consumer+WKT.swift
```diff
@@ -0,0 +1,322 @@
...
+                throw WKTError.coordinatesParsingFailed(values)
```
kried, line 228: Would it be possible to add tests for incorrect WKT input strings? E.g. wrong numbers of elements, incorrect type, etc.

Sources/Turf/Consumer/Consumer+WKT.swift
```diff
@@ -0,0 +1,322 @@
...
+    } ()
+    
+    private lazy var wktConsumer: Consumer<WKTLabel> = .interleaved([
+        .spaced(object)
+    ], .discard(","))
+    
+    // Transform
+    private let wktTransform: Consumer<WKTLabel>.Transform = { label, values in
+        switch label {
+        case .empty:
```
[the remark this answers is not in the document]
Udumft: I've re-checked, XCode things it's fine

Sources/Turf/Consumer/Consumer+WKT.swift
```diff
@@ -0,0 +1,322 @@
+import Foundation
+#if !os(Linux)
+import CoreLocation
+#endif
+
+
+struct WKTParser {
```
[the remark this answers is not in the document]
Udumft: Done

Sources/Turf/Consumer/Consumer+WKT.swift
```diff
@@ -0,0 +1,322 @@
...
+    init() {}
+    
+    mutating func parse(_ input: String) throws -> Any {
+        let match = try wktConsumer.match(input.uppercased())
+        return try match.transform(wktTransform)!
```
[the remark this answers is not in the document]
Udumft: I've missed that one, thanks.

Sources/Turf/Consumer/Consumer+WKT.swift
Udumft: Appended

Sources/Turf/Consumer/Consumer+WKT.swift
```diff
@@ -0,0 +1,327 @@
...
+            return coords
+        case .point:
+            guard values.count > 1 else { return nil }
+            guard let coords = values[1] as? LocationCoordinate2D else {
+                throw WKTError.coordinatesParsingFailed(values)
+            }
+            return Point(coords)
+        case .multiPoint:
+            guard values.count > 1 else { return nil }
+            let coords = values[1..<values.endIndex].compactMap { $0 as? LocationCoordinate2D }
```
1ec5, line 260: This is correct, but we can avoid a range here and in similar situations for additional safety assurance:

```suggestion
            let coords = values.suffix(from: 1).compactMap { $0 as? LocationCoordinate2D }
```

Sources/Turf/Consumer/Consumer+WKT.swift
```diff
@@ -0,0 +1,327 @@
...
+        case .object:
+            return values.first
+        }
+    }
+}
+
+
+extension Consumer where Label == WKTParser.WKTLabel {
+    private static var space: Consumer<WKTParser.WKTLabel> {
+        return .discard(.zeroOrMore(.character(in: " \t\n\r")))
```
1ec5, line 305: Are these the only whitespace characters to consider, or should we consider all of `CharacterSet.whitespacesAndNewlines`?

Sources/Turf/WKT.swift
```diff
@@ -0,0 +1,222 @@
...
+    
+    public init?(fromWKT wkt: String) {
+        do {
+            let data = try WKTParser.parse(wkt)
+            guard let object = (data as? [MultiPolygon])?.first else {
+                return nil
+            }
+            self = object
+        } catch {
+            print("error: Failed to init 'MultiPolygon' from WKT with error: \(error)")
```
1ec5, line 119: Should there be a throwable initializer in order to pass the error further up the stack? If the error is ignorable, then perhaps we shouldn’t spew it to the console.

Sources/Turf/WKT.swift
```diff
@@ -0,0 +1,222 @@
...
+        case .polygon(let geometry):
+            return geometry.WKTString
+        case .multiPoint(let geometry):
+            return geometry.WKTString
+        case .multiLineString(let geometry):
+            return geometry.WKTString
+        case .multiPolygon(let geometry):
+            return geometry.WKTString
+        case .geometryCollection(let geometry):
+            return geometry.WKTString
```
1ec5, line 141: We didn’t end up creating a bottleneck method like this for `GeometryConvertible.geometry`, so should we do it for `WKTCodable.WKTString`? Or is the rationale for this bottleneck method that decoding to WKT shouldn’t require using a `Decoder`?

Sources/Turf/WKT.swift
```diff
@@ -0,0 +1,222 @@
+import Foundation
+
+/**
+ Entity which can be encoded to and decoded from 'Well Known Text'.
+ */
+public protocol WKTCodable {
+    var WKTString: String { get }
```
1ec5, line 7: Rename to `wktString`. [Swift case conventions](https://www.swift.org/documentation/api-design-guidelines/#follow-case-conventions) lowercase acronyms when they occur at the beginning of a non-type symbol name.

Sources/Turf/WKT.swift
```diff
@@ -0,0 +1,222 @@
+import Foundation
+
+/**
+ Entity which can be encoded to and decoded from 'Well Known Text'.
+ */
+public protocol WKTCodable {
```
1ec5, line 6: The name of this protocol implies that it’s somehow related to Codable, but it doesn’t appear to be. Instead, it looks very similar to protocols such as GeometryConvertible and GeoJSONObjectConvertible. Consider renaming this protocol to WKTConvertible.

Sources/Turf/Consumer/Consumer.swift
```diff
@@ -0,0 +1,887 @@
...
+//  Consumer.swift
+//  Consumer
+//
+//  Created by Nick Lockwood on 01/03/2018.
+//  Copyright © 2018 Nick Lockwood. All rights reserved.
+//
+//  Distributed under the permissive MIT license
+//  Get the latest version from here:
+//
+//  https://github.com/nicklockwood/Consumer
```
1ec5, line 11: Can you elaborate on the choice to vendor this library instead of depending on it using a package manager? If we need to avoid an additional third-party dependency, is it possible to accomplish the same functionality using the built-in `Scanner` type from Foundation?

Sources/Turf/WKT.swift
Udumft: Didn't want to silently fail, so put a log message. But in the end Error is preferable way, so refactored parser a bit for that purpose. Thanks for flagging it

Sources/Turf/WKT.swift
Udumft: `Decoder` is indeed not a preferable way here (IMHO) for several reasons. There is no clear path or guide on how to write custom Coders and Decoders. Related publicly visible protocols are not meant to be guide on that process, and majority of logic for existing JSONDecoder or PropertyListDecoder are hidden. Also, this would potentially mean re-using `encode/decode` methods, but since WKT and JSON formats are too different (in terms of syntax and data content) this would overcomplicate the methods. Plus, there should be some trick required to distinguish which type of coder is used for each particular call.
And at last our use-case is actually not too complex, so I decided we could get away with simple method which would convert object to a string. There is no wild nesting or huge variety of types and data that could be inside geometries, so full scale Coder may be an overkill.

Sources/Turf/Consumer/Consumer.swift
Udumft: My motivation to include the source directly is that this lib is basically a single file, it will not bloat our repo and we most likely will never need to update it's version. Linking through all 3 managers seemed like unnecessary burden.
Built-in features like `Scanner` or `RegEx` did not fit due to minimal OS versions incompatibility. For example majority of required `Scanner` methods require iOS 13.

1ec5: > For example majority of required `Scanner` methods require iOS 13.

`Scanner` is equally capable in iOS 12. All that’s new in iOS 13 are the method overloads for returning an optional type. For example, [`scanFloat(_:)`](https://developer.apple.com/documentation/foundation/scanner/1411677-scanfloat) was introduced in iOS 2.0 as a C-style method that takes an `UnsafeMutablePointer` and returns a Boolean. iOS 13 introduced [`scanFloat(representation:)](https://developer.apple.com/documentation/foundation/scanner/3200284-scanfloat), which returns a `Float?`. The new methods are certainly more convenient, but it should be possible to accomplish the same functionality using the old C-style methods.

Since the current implementation does work fine, we can land this feature based on Consumer. However, we should make its public types private so that downstream packages can continue to depend on Consumer without any conflicts. Consumer is now a fairly large portion of this codebase, so let’s open a separate issue to track replacing Consumer with Scanner as tail work.

Sources/Turf/WKT.swift
1ec5: I agree about avoiding Coder, although I meant to ask if we really needed `Geometry.WKTString` as a bottleneck method. On second thought, this method is fine. 👍 Unlike `GeometryConvertible.geometry`, `WKTConvertible.wktString` always returns a `String`, so there’s no benefit to making callers perform this switch themselves. Also, the `Geometry` type only serves a fixed number of geometry types; any other `GeometryConvertible` type would be irrelevant here.

Sources/Turf/WKT.swift
```diff
@@ -0,0 +1,151 @@
+import Foundation
+
+/**
+ Entity which can be converted to and from 'Well Known Text'.
+ */
+public protocol WKTConvertible {
+    var wktString: String { get }
+    init(fromWKT wkt: String) throws
```
1ec5, line 8: The “from” in the argument label is unnecessary. There does still need to be a `wkt` label because not every string is a WKT string:

```swift
init(wkt: String) throws
```

Sources/Turf/WKT.swift
```diff
@@ -0,0 +1,151 @@
+import Foundation
+
+/**
+ Entity which can be converted to and from 'Well Known Text'.
+ */
+public protocol WKTConvertible {
+    var wktString: String { get }
```
1ec5, line 7: Is the “string” redundant? Should it be just `wkt`?

README.md
```diff
@@ -135,3 +135,24 @@ print(json)
...
+## Well Known Text (WKT)
+
+turf-swift contains a minimal WKT encoding/decoding support for geometries implementing `WKTCodable` protocol.
+
+```swift
+let wktString = "POINT(123.53 -12.12)"
+
+// Decoding is done using an init method
+let point = Point(fromWKT: wktString)
+let geometry = Geometry(fromWKT: wktString)
```
1ec5, line 148: This initializer is throwable, so stick a `try?` in here.

1ec5 (approved):
Some additional suggested refactors, not strictly necessary, but otherwise good to go.

PR #205, opened 2022-12-22
Enabling automatic test discovery on Linux

Udumft:
PR updates swift version used on Travis CI linux job which enables automatic test discovery on Linux platform. Unit tests and SPM target corrected with found errors. 
Had to update Linux distributive from `Trusty` to `Xenial` since Ubuntu 14.04 does not come with Swift 5.4 support, and minimal supported version is Ubuntu 16.04 (as found [here](https://www.swift.org/download/)) which is represented with `Xenial` dist on [Travis](https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment).

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.5
+// swift-tools-version:5.4
 // The swift-tools-version declares the minimum version of Swift required to build this package.
 
 import PackageDescription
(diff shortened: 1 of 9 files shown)


Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:5.0
+// swift-tools-version:5.4
```
MaximAlien, line 1: I recommend testing this change on Mapbox Directions directly to make sure there are no regressions.

Udumft: I tried MapboxDirections with this dependency and it resolved and compiled correctly.

PR #213, opened 2024-01-16
Support visionOS target

persidskiy:
This PR supports visionOS target. 

It also bumps the **minimal xcode version to 14.1**. This is required to use `.custom("visionOS", versionString: "1.0")` in spm, which is supported starting from swift 5.6. However, I decided it's safe to bump the minimal swift version to 5.7, as Xcode 14.1 is the minimal supported version for AppStore distribution.

Also, added CI test runs on iOS and visionOS simulators.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,6 +26,23 @@ jobs:
       - run:
           name: "Test"
           command: swift test
+  test_simulator:
+    parameters:
+      destination:
+        type: string
+    macos:
+      xcode: 15.2.0
+    resource_class: macos.m1.medium.gen1
+    steps:
+      - checkout
+      - run:
+          name: "Test << parameters.destination >>"
+          command: |
+            xcodebuild test \
+              -scheme Turf \
+              -destination "<< parameters.destination >>" \
+              -derivedDataPath build
+
   build_for_library_evolution:
     macos:
       xcode: 15.1.0
@@ -47,4 +64,10 @@ workflows:
       - build_and_test_macos:
           matrix:
             parameters:
-              xcode_version: [13.4.1, 14.3.1, 15.1.0]
+              xcode_version: [14.3.1, 15.1.0]
+      - test_simulator:
+          matrix:
+            parameters:
+              destination:
+                - "platform=visionOS Simulator,OS=1.0,name=Apple Vision Pro"
+                - "platform=iOS Simulator,OS=17.2,name=iPhone 15"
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -26,6 +26,23 @@ jobs:
       - run:
           name: "Test"
           command: swift test
+  test_simulator:
+    parameters:
+      destination:
+        type: string
+    macos:
+      xcode: 15.2.0
+    resource_class: macos.m1.medium.gen1
```
persidskiy, line 35: visionOS simulator is only supported on M1

Turf.podspec
```diff
@@ -24,6 +24,9 @@ Pod::Spec.new do |s|
   s.osx.deployment_target = "10.13"
   s.tvos.deployment_target = "11.0"
   s.watchos.deployment_target = "4.0"
+  # CocoaPods doesn't support releasing of visionOS pods yet, need to wait for v1.15.0 release of CocoaPods
+  # with this fix https://github.com/CocoaPods/CocoaPods/pull/12159.
+  # s.visionos.deployment_target = "1.0"
```
persidskiy, line 29: Having this fails the `pod spec lint` with the following error:
```
    - ERROR | [visionOS] unknown: Encountered an unknown error (Could not find a `visionos` simulator (valid values: ios, tvos, watchos, xros). Ensure that Xcode -> Window -> Devices has at least one `visionos` simulator listed or otherwise add one.) during validation.
```

I think it's fine not to support VisionOS for CocoaPods distribution for a while.

Package.swift
```diff
@@ -1,12 +1,12 @@
-// swift-tools-version:5.5
+// swift-tools-version:5.7
```
azarovalex, line 1: We could leave this file unmodified and create a separate one named `Package@swift-5.7.swift`, but it's probably okay to drop Xcode 13 support already.

persidskiy: Yeah, that was initial idea, but then it's a duplicate and `swift build` was failing. So I decided it worth to update xcode requirement, as it xcode 13 is not supported in appstore anyway.

Turf.podspec
OdNairy: Can we use the `main` branch of CocoaPods for releases for now? Not sure if it will break integration for customers with pre-1.15 CocoaPods

OdNairy (approved):
LGTM
Makes sense to bump min Xcode requirement

Turf.podspec
persidskiy: They say it doesn't work still https://github.com/CocoaPods/CocoaPods/pull/12159#issuecomment-1897118199

If nobody fixes it and I have time, I'll try to fix it myself.

PR #219, opened 2024-08-23
Remove public RawRepresentable conformances

OdNairy:
* Remove public `RawRepresentable` conformance from the Array and Dictionary

MAPSIOS-1148


persidskiy (approved):
LGTM

PR #220, opened 2024-08-27
Add convenience accessors to enum values

persidskiy:
This PR adds several extensions that allow enum value to be easily red and chained.

For example,
```swift
aFeature.geometry.point?.map(\.foo)
```

```swift
if let fooString = aFeature.properties.object?["foo"].string {
}
```

Doing the same with the `if let .case(foo) = bar` construction is usually cumbersome and doesn't support chaining.


OdNairy (approved):
Looks like a great addition, thanks!

PR #243, opened 2024-10-26
Establish release process for Turf binary distribution

aleksproger:
- Distribute Turf as xcframework in SwiftPM and CocoaPods
- Create release and pre-release scripts to fully automate release process
- Build from source using xcodebuild instead of SwiftPM
- Delete jobs for building on Linux and separete job for building with library evolution
- Add create xcframework job on PRs

Issues solved:
- Clients that depend on source-based Turf and MapboxSDK stack won't receive two copies ob library in SwiftPM, compatibility will be ensured by SwiftPM


1ec5:
Out of curiosity, could you speak to why this library is no longer being distributed as source code and no longer supporting Linux?

aleksproger:
@1ec5, Turf will become a dependency of other Mapbox library deeper in Mapbox SDK stack and this library is distributed as binary only dynamic framework, which means that Turf needs to be distributed as binary as well. In order to eliminate the issues when client may depend on Mapbox SDK stack where binary dependency exists and also on source-based Turf it was decided to switch to binary distribution of Turf and allow SwiftPM properly resolve compatibility issues.
Technically, the support for Linux is not dropped, however it was only in SwiftPM and now we don't build with SwiftPM, but use .xcframework binary instead.

aokj4ck (approved):

OdNairy (approved):
LGTM
I have some concerns about taking away source code from the SPM but since we didn't find a way to keep it along with the Common depends on the turf we have to go that way.

I see that we can't use `pod spec lint` anymore but I think we should use the `pod lib lint` then instead (it will use local files instead of url).

